PR #4, opened 2019-10-28
Migrated content from legacy site md files

jessicaschilling:
Heads up: Lots of internal links will be broken; leaving as-is so we can swarm on fixes as a larger group
Ref https://github.com/ipfs/docs/issues/345


jessicaschilling:
@cwaring and @ericronne -- Heads up that I've made a first pass at importing existing legacy Markdown files into the beta site. Lots of internal links will be broken, and a variety of other issues ... as we've discussed, we'll need to comb through all of the content for a first-pass cleanup for formatting and to fix all links. I've elected to merge in this state, though, so that we can swarm as a larger group on going over the documents if need be. Thanks!

ericronne:
Will check it. Stoked!! Can almost taste it 😋

PR #13, opened 2019-11-01
Feat add corecourses

jessicaschilling:
Incorporated the four core courses from IPFS Camp:
- Courses A and D are broad enough to touch multiple "essentials" topics, so added tip callouts to them in their respective topic pages, plus prompts in the top-level essentials page
- Courses B and C added to nav in the "How-tos" section

Ref https://github.com/ipfs/docs/issues/187


jessicaschilling:
@cwaring -- This isn't urgent by any means, but on Monday, can you please look into why my checks failed? Looks like there were some earlier failures with the nav amends earlier, as well; not sure if you're aware of it. Thank you!

cwaring:
@jessicaschilling I've rebased against master and the CI issues are fixed now. This looks good to me so I'll merge it down 🙌

PR #28, opened 2019-12-03
feat: content status with related resources

cwaring:
This PR implements https://github.com/ipfs/docs/issues/342.

It registers a new global component that can be used in any markdown file, simply add the following line:

```jsx
<ContentStatus title="**optional**" />
```

You can also use front matter data to configure the display and add related items.

```yml
issueNum: 333 # used to connect to the related issue for the content
related: # nested key, value items rendered as links
  an interesting article: https://ipfs.io
  another interesting article: https://ipfs.io
```

cs-dt
cs-mb

The interaction logic for the voting mechanism will come in a later PR.


jessicaschilling:
LGTM 😍

cwaring:
Thanks @ericronne, I agree but ideally tweaks 2-6 should not be not part of this PR because we need to setup global styles for them to retain consistency and easy of maintenance - otherwise the app will soon get littered with unique pixel values. I would suggest that we address it in the CSS edits branch so it can be become an inherited style.

I'll apply a quick patch in this branch for a visual ref so we can migrate it across later, including the button style etc :)

ericronne:
Yes i suspected as much.  ✊

PR #35, opened 2019-12-04
On beta, adds callout for feedback/beta signup/equivalent legacy pages

jessicaschilling:
For review by @cwaring and @ericronne on the understanding that this isn't exactly as [visually proposed ](https://github.com/ipfs/docs/issues/341#issuecomment-553577494) in the original content thread and could likely be more efficiently built, but it gets the job done for something that will only be a temporary part of the beta. Y'all can discuss how much it's worth tweaking. 😉 This completes the work that allows us to close https://github.com/ipfs/docs/issues/341.
- Adds "this is a beta" callout to the bottom of all content pages (based visually on existing "tip" styles)
- Includes a link to the equivalent legacy page when one exists
- Includes a link to submit feedback (as a generic documentation issue on GitHub in the legacy repo)
- Includes a link to the beta test signup form (using same logic as in the content stub pages)

Example below; note that the "View this page on legacy site" link doesn't appear if there's no equivalent page on legacy.
image


jessicaschilling:
Leaving this PR open until @cwaring determines if/whether/how this should be incorporated into a site footer.

ericronne:
My commit moves the feature suggestion to this legacy callout, but requires additional formatting by Sir @cwaring. 🙏 

## As of this commit …
image

## Desired design …
image

## To do …
- [ ] Add vertical spacing above the callout (should match spacing between major page sections)
- [ ] Split the four links into two columns on desktop
- [ ] Change the "View this page on legacy site" to regular (thinner) weight font

Related: [This issue](https://github.com/ipfs/docs/issues/406) to remove the feature request from the header

PR #39, opened 2019-12-09
Relabels suggest a 'feature' to 'topic'

ericronne:
Soon we'll suss out a better experience path for user suggestions re content and feature improvements, but meanwhile, this hopefully this change will elicit some good feedback.


ericronne:
OBE, modifying approach to unify the suggestion experience …
cf https://github.com/ipfs/ipfs-docs-v2/pull/35#issue-349070128

PR #43, opened 2019-12-16
Unlinks words and adjusts padding on two bottom-of-pg

ericronne:
- Unlinks "on GitHub" in the feedback tool, which serves to visually separate the two links in this widget, so users can more easily distinguish them at a glance:
image

- Adds padding to the left side of the feedback tool and the legacy callout, in order to align them with the page content in desktop view:
image

@cwaring if you have time to further adjust (or ideally, 🍐code with me), these additional tweaks would improve the layout on the legacy callout. If not, we can merge and address later …
- [x] Reduce the width of the green container, so that it right aligns with the page content. (Ideally the green would only blow out to the full width on mobile.)
- [x] Narrow the two columns accordingly (so "Give general feedback" and "Become a tester" shift leftward)
- [x] Add left padding to legacy callout on tablet view, as well

thx!

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -41,7 +41,7 @@ module.exports = {
         label: 'English',
         selectText: 'Languages',
         ariaLabel: 'Select language',
-        editLinkText: 'Edit this page on GitHub',
+        editLinkText: 'Edit this page',
         lastUpdated: 'Last Updated',
         serviceWorker: {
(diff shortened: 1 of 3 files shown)


jessicaschilling:
Good call on the links. I'd **strongly** suggest merging this and letting @cwaring work on higher-urgency issues, though; he's got far too much on his plate until the end of the year. Thank you!

PR #46, opened 2019-12-17
feat: support ipfs

hugomrdias:
This PR adds support for hosting in IPFS.

Steps:
`npx vuepress build docs`
`all-relative ./docs/.vuepress/dist`
`ipfs add ./docs/.vuepress/dist`
Enjoy.

Theres still a couple of assests in the public folder that all-relative doesn't handle but that should be easy to fix.

Working example https://ipfs.io/ipfs/Qmb3bTUHXuxcuU9VWB5yG99Usa826kw4ZagyZJ5DucSZo3/

### DIFF
--- a/docs/.vuepress/public-path.js
+++ b/docs/.vuepress/public-path.js
@@ -1,0 +1,12 @@
+'use strict'
+
+;(function() {
+  var ipfsPathRegExp = /^(\/(?:ipfs|ipns)\/[^/]+)/
+  var ipfsPathPrefix =
+    (window.location.pathname.match(ipfsPathRegExp) || [])[1] || ''
+
+  __webpack_public_path__ = ipfsPathPrefix + '/'
+  if (typeof window !== 'undefined') {
+    window.__VUEPRESS_ROUTER_BASE__ = ipfsPathPrefix + '/'
+  }
+})()
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,44 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/node:10.15.1-browsers
+        environment:
+          NO_SANDBOX: true
+    steps:
+      - checkout
+      - run:
+          command: npm run build:docs:ipfs
+      - persist_to_workspace:
+          root: .
+          paths:
+            - ./docs/.vuepress/dist
+
+  deploy:
+    docker:
+      - image: olizilla/ipfs-dns-deploy:latest
+        environment:
+          DOMAIN: docs-beta.ipfs.io
+          BUILD_DIR: ./docs/.vuepress/dist
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+      - run:
+          name: Deploy website to IPFS
+          command: |
+            pin_name="$DOMAIN build $CIRCLE_BUILD_NUMBER"
+            hash=$(pin-to-cluster.sh "$pin_name" /tmp/workspace/$BUILD_DIR)
+            echo "Website added to IPFS: https://ipfs.io/ipfs/$hash"
+            # Update DNSlink for prod or dev domain
+            if [ "$CIRCLE_BRANCH" == "master" ] ; then
+              dnslink-dnsimple -d $DOMAIN -r _dnslink -l /ipfs/$hash
+            fi
+workflows:
+  version: 2
+  build-deploy:
+    jobs:
+      - build
+      - deploy:
+          context: ipfs-dns-deploy
+          requires:
+            - build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,17 +8,17 @@ jobs:
     steps:
       - checkout
       - run:
-        name: update-npm
-        command: 'sudo npm install -g npm@latest'
-    - restore_cache:
-        key: dependency-cache-{{ checksum "package-lock.json" }}
-    - run:
-        name: install-npm
-        command: npm install
-    - save_cache:
-        key: dependency-cache-{{ checksum "package-lock.json" }}
-        paths:
-          - ./node_modules
+          name: update-npm
+          command: 'sudo npm install -g npm@latest'
+      - restore_cache:
+          key: dependency-cache-{{ checksum "package-lock.json" }}
+      - run:
+          name: install-npm
+          command: npm install
+      - save_cache:
+          key: dependency-cache-{{ checksum "package-lock.json" }}
+          paths:
+            - ./node_modules
       - run:
           name: build-docs-ipfs
           command: npm run docs:build:ipfs
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,19 @@ jobs:
     steps:
       - checkout
       - run:
+        name: update-npm
+        command: 'sudo npm install -g npm@latest'
+    - restore_cache:
+        key: dependency-cache-{{ checksum "package-lock.json" }}
+    - run:
+        name: install-npm
+        command: npm install
+    - save_cache:
+        key: dependency-cache-{{ checksum "package-lock.json" }}
+        paths:
+          - ./node_modules
+      - run:
+          name: build-docs-ipfs
           command: npm run docs:build:ipfs
       - persist_to_workspace:
           root: .
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
     steps:
       - checkout
       - run:
-          command: npm run build:docs:ipfs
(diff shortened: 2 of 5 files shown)


hugomrdias:
thanks @olizilla for `all-relative`

docs/.vuepress/public-path.js
```diff
@@ -0,0 +1,12 @@
+'use strict'
+
+;(function() {
+  var ipfsPathRegExp = /^(\/(?:ipfs|ipns)\/[^/]+)/
+  var ipfsPathPrefix =
+    (window.location.pathname.match(ipfsPathRegExp) || [])[1] || ''
+
+  __webpack_public_path__ = ipfsPathPrefix + '/'
+  if (typeof window !== 'undefined') {
+    window.__VUEPRESS_ROUTER_BASE__ = ipfsPathPrefix + '/'
```
cwaring: nice! this var is the secret ingredient, where did you find it?

hugomrdias: vuepress bootstrap code.
but `__webpack_public_path__ = ipfsPathPrefix + '/'` this one is even more important without it the router would even load

hugomrdias: https://github.com/vuejs/vuepress/blob/master/packages/%40vuepress/core/lib/client/app.js#L63

cwaring:
Yes, I'm so happy we didn't have to regex the application source in the end! Do you think we could wrap this up neatly into a plugin by using https://vuepress.vuejs.org/plugin/option-api.html#chainwebpack to handle the path injection script?

docs/.vuepress/public-path.js
cwaring: I knew there was a `__webpack_public_path__` lurking about but I missed this one! Great work detective 🕵️‍♂️

hugomrdias:
> Yes, I'm so happy we didn't have to regex the application source in the end! Do you think we could wrap this up neatly into a plugin by using [vuepress.vuejs.org/plugin/option-api.html#chainwebpack](https://vuepress.vuejs.org/plugin/option-api.html#chainwebpack) to handle the path injection script?

yes we should be able to build a plugin to do all this in one go, just need to figure out the last missing assets either but improving all-relative or in the plugin itself.

docs/.vuepress/public-path.js
hugomrdias: yeah @satazor put me in the right direction with that one i just did the last mile

olizilla:
@hugomrdias I'm guessing it would be helpful if I actually published all-relative to npm!

olizilla:
It a thing https://www.npmjs.com/package/all-relative

docs/.vuepress/public-path.js
cwaring: the highest of fives to @satazor! It also makes me wonder if we could make a more generic IPFS plugin for webpack compiled apps using some of these framework config vars. we'd need to do detection for router logic but it doesn't seem impossible now... 🤔

cwaring:
thanks @olizilla, so excited to see this all come together!

docs/.vuepress/public-path.js
satazor: Thanks! I also thought about that in the past @cwaring and it’s doable. The route logic detection could also be pluggable, meaning we could have addons/plugins typical frameworks such as Gatsby and Vue press.

I guess it would take some investment time but it would pay off later.

cwaring: nice, if we are delaying the adoption of origin isolation for a while then this might be a worthy endeavour to the community. I see this question crop up all the time.

cwaring:
I've fixed the relative css path issues upstream in `all-relative` and enabled the circleci build config. So when @olizilla gets a few min to cut a new release we should be good to merge this and update the DNS for full IPFS support! 🎉

hugomrdias:
I think some <head> loads were failing also like manifest etc

cwaring:
Yes I've noticed that too but they are not critical for launch and I don't think `all-relative` should touch them by default. I'd only like to enable PWA functionality from the root domain so I can pick that up in a future update :)

samajammin:
Hey folks! Thanks for providing this as a reference!

I'm trying to add IPFS support on ethereum.org (https://github.com/ethereum/ethereum-org-website/pull/751). I've copied your approach - it's up & running locally! This seems to have solved the issue with relative links. The only issue I see now is broken images from the /public dir. @hugomrdias - you mentioned that is your PR description. Did you find a solution to that?

Thanks again.

hugomrdias:
ill take a look

samajammin:
Awesome, thanks for the guidance!

So I now have the site pinned here: https://gateway.pinata.cloud/ipfs/QmZxrnb4P3RHcHpF7mTRzQ5x2vF1b5vZtWR23Ef5MsAcV9/

Looking at my build output in /docs/.vuepress/dist, the images seem to be relative links as intended but I'm seeing some inconsistent results on the IPFS gateway.

e.g. sometimes I load [this page](https://gateway.pinata.cloud/ipfs/QmZxrnb4P3RHcHpF7mTRzQ5x2vF1b5vZtWR23Ef5MsAcV9/build) & the images appear, with DOM inspection showing relative links:

Screen Shot 2020-02-19 at 8 33 24 AM

Sometimes I reload the page & the images are broken. The source shows they _are not_ relative links:
Screen Shot 2020-02-19 at 8 33 04 AM

Do you see this as well? Any idea what may be causing the inconsistency here?

samajammin:
One more item I'm noticing - the mp4 file on our homepage is not referenced correctly:

Screen Shot 2020-02-19 at 8 51 34 AM

Looking at the build output HTML:
```
<video id="hero-video" alt="Ethereum.org - Light" width="380" height="380" src="/ethereum-hero-light.mp4" playsinline autoplay loop muted class="mx-auto inline-block" data-v-5d8b6e4d=""></video>
```
I believe the src should be `src="./ethereum-hero-light.mp4"` vs. `src="/ethereum-hero-light.mp4"`. Updating this manually in the browser gets it working. I'm not sure why this particular video element / mp4 file is having issues, [it looks like all-relative searches for all nodes w/ a `src` property](https://github.com/olizilla/all-relative/blob/master/process-html.js#L15), which would still apply to this?

cwaring:
Interesting, this isn't something we've seen on the docs project yet. It seems that VuePress is outputting absolute paths when the page/components re-render after routing. I haven't had chance to dig any deeper into your codebase but I wonder if it's something to do with how those individual assets are being included within the components?

A couple of things I'd try next:
+ Is it possible to make those asset inclusions baseurl aware?
+ What happens if you switch inline images to display using a CSS background instead?

Let me know if either of these routes work and I'll aim to dig in as soon as I can.

cwaring:
@samajammin curiosity got the better of me and I think I've found [the culprit](https://github.com/ethereum/ethereum-org-website/blob/8b7b2b0c50ad30c3ed5d3676667dc96111574bb1/docs/.vuepress/theme/components/Hero.vue#L82)! 

If you are injecting asset paths into the markup dynamically (via strings) you will have to ensure they are relative. The easiest way to do this is to hook into `window.__VUEPRESS_ROUTER_BASE__` and if it exists prepend it to the root path.

A meta function like this should work (⚠️ not tested):

```js
const withBasePath = path => window && window.__VUEPRESS_ROUTER_BASE___ ? window.__VUEPRESS_ROUTER_BASE__ + path : path || '/' + path;

withBasePath('ethereum-hero-dark.mp4')
// => "/ipfs/QmZxrnb4P3RHcHpF7mTRzQ5x2vF1b5vZtWR23Ef5MsAcV9/ethereum-hero-dark.mp4"
```

Let me know how it goes, I likely add a utility like this to the final plugin 🤙🏻

cwaring:
Or perhaps you could modify `public-path.js` script to inject a [base url tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) into the header once if it detects a window object, that should make all path urls work correctly even if they are injected at runtime. This method would also need to be tested to see how it interacts with the router links.

samajammin:
Thanks for the ideas @cwaring! I appreciate you taking the time to look into this.

I tried your suggestion to create a `withBasePath` helper but unfortunately that did not solve the issue. I suspect this is because the site build happens in node, where `window.__VUEPRESS_ROUTER_BASE___` is not available. As a result, those images still do not have the relative path needed.

When I have a chance (hopefully this week) I will keep experimenting but if you have any more ideas or if there's something you think I'm missing, please let me know.

samajammin:
Here's the PR where I tried that update in case you'd like to experiment on your own:
https://github.com/ethereum/ethereum-org-website/pull/751

PR #47, opened 2019-12-18
feat: sitewide metadata and seo

cwaring:
Now it is as easy as adding a single `description` key to the front matter of a page to add rich meta previews. You can override any individual items on a page by page basis by appending the items to the front matter.

Implements: https://github.com/ipfs/docs/issues/343


jessicaschilling:
Doo eet!

cwaring:
I've tried using `vuepress-plugin-autometa` as an alternative, it is better in some ways however with the current version there isn't a way to set a default fallback meta card image. I suggest we merge this for now and revisit before public launch and allow the robots in.

cwaring:
closing in favour of: https://github.com/ipfs/ipfs-docs-v2/pull/36

PR #50, opened 2019-12-22
fix: add canonical link

lidel:
This PR adds `<link rel="canonical"` header ([rfc6596](https://tools.ietf.org/html/rfc6596), [Wikipedia](https://en.wikipedia.org/wiki/Canonical_link_element)) to deduplicate SEO across all copies loaded from various public IPFS gateways.

Prior Art:
- https://github.com/ipfs/distributed-wikipedia-mirror/issues/65
- https://github.com/vuejs/vuepress/issues/894

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -364,7 +364,7 @@ module.exports = {
       {
         // add <link rel="canonical" header (https://tools.ietf.org/html/rfc6596)
         // to deduplicate SEO across all copies loaded from various public gateways
-        baseURL: 'https://docs.ipfs.io'
+        baseURL: 'https://docs-beta.ipfs.io'
       }
     ]
   ],
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -358,6 +358,14 @@ module.exports = {
           add('keywords', $site.themeConfig.keywords)
         }
       }
+    ],
+    [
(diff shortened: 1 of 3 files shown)


jessicaschilling:
@cwaring -- I'd merge this, but am not sure what's up with the CircleCI build error situation. Can you please look when you get back from holidays? Thank you!

docs/.vuepress/config.js
```diff
@@ -358,6 +358,14 @@ module.exports = {
           add('keywords', $site.themeConfig.keywords)
         }
       }
+    ],
+    [
+      'vuepress-plugin-canonical',
+      {
+        // add <link rel="canonical" header (https://tools.ietf.org/html/rfc6596)
+        // to deduplicate SEO across all copies loaded from various public gateways
+        baseURL: 'https://docs.ipfs.io'
```
cwaring: I assume this should reflect the primary site access point, such as `docs-beta.ipfs.io` then `docs.ipfs.io` once the original site is deprecated?

For the meantime I've added a rule to stop robots indexing the site while it's in beta, however we might decide to change that once we've rearranged a few sections.

lidel: Yes, if we want to deploy this to `docs-beta.ipfs.io` first, then we should use it:

```suggestion
        baseURL: 'https://docs-beta.ipfs.io'
```

PR #66, opened 2020-01-16
fix: IFPS → IPFS

lidel:
:^)


cwaring:
We need a linter rule for this one 😅

lidel:
@hugomrdias seriously tho, is this something we can add to `aegir lint` ?

hugomrdias:
Haha it's going to be hard to find a rule to cover this but hey everything is possible

PR #74, opened 2020-01-23
fix: smooth scrolling

cwaring:
fixes: https://github.com/ipfs/docs/issues/429

@johnnymatthews let me know if this patch works. It means that FF will lose smooth scrolling but I feel that is the best workaround considering the inconstancies here.


johnnymatthews:
Yeah this words perfectly fine @cwaring, nice one.

PR #99, opened 2020-02-12
Docs contribution section

johnnymatthews:
The majority of contribution content is contained within this PR. The only thing that isn't here is the _Code persona_, which still needs a bunch of discussion before we commit to anything.


johnnymatthews:
I'm not massively happy about how the [Ways to contribute](https://docs-beta.ipfs.io/community/contribute/ways-to-contribute/) link shows up in the sidebar. It doesn't need to fit into another sub-category (like _Write the docs_), but it'd be nice if it stood out a bit more.

Screen Shot 2020-02-12 at 2 16 31 PM

Anyway, it'll do for now.

jessicaschilling:
@johnnymatthews This nav structure omits the IRC, forums, and social media info from the left-hand nav; we need to put them all back in that left nav or they disappear from public consciousness.
I've taken a pass at addressing this in https://github.com/ipfs/ipfs-docs-v2/pull/103, but feel free to tweak further.

johnnymatthews:
So instead of me pushing broken stuff into master, then having to roll stuff back when I break everything, is there a way to preview my working-branch changes somewhere @cwaring. That way we can look at the changes on a pre-production server/service, have a discussion about what's going on, and then merge the correct things into `master`.

The past few merges I've made into `master` either end up braking something else down the line, or I'm missing conventions that have been discussed somewhere at some point that I'm not aware of.

jessicaschilling:
@johnnymatthews - Expand "show all checks" in a PR and there's a link to a live preview. (Note this is only for internal folks, unfortunately ... only folks in the org trigger the entire CI workflow.)

image

For larger-scale work, it's always best to request peer review before merging, in any case.

PR #100, opened 2020-02-12
Update grammar-formatting-and-style.md

jessicaschilling:
Broken images, broken links, copy editing for grammar, consistency, typos
fyi @johnnymatthews

### DIFF
--- a/docs/community/contribute/grammar-formatting-and-style.md
+++ b/docs/community/contribute/grammar-formatting-and-style.md
@@ -4,11 +4,11 @@ This page details the syntax and formatting rules for writing IPFS documentation
 
 ## Grammar and spelling
 
-Here are some language specific rules that the IPFS documentation follows. Most of these rules are turned on by default in Grammarly.
+Here are some language-specific rules that the IPFS documentation follows. If you use a writing service like [Grammarly](https://www.grammarly.com/), most of these rules are turned on by default.
 
 ### American English
 
-While IPFS is a global project, the fact is that American English is the most commonly used _style_ of English used today. With that in mind, when writing content from the IPFS project, use American English spelling. The basic rules are:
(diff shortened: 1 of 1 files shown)


cwaring:
PR 💯!
Congrats you win an emoji 🎁

johnnymatthews:
Didn't realize we had to hard link images though `raw.githubusercontent.com`. Has that always been the case, or is this a new thing?

jessicaschilling:
It's the only way I've found to be able to reliably ensure images display in both GitHub and the rendered VuePress version.

cwaring:
FWIW: I would favour relative links otherwise the images won't get hashed and optimised during the build pipeline and also won't be visible offline or when fetched via IPFS. Relative images _should_ display in Github too? If not we should look into that separately.

jessicaschilling:
In the original version, the relative links displayed in GitHub, but not the VuePress version.

johnnymatthews:
Just checking I'm on the same page here, what would a relative link look like for [line 46](https://github.com/ipfs/ipfs-docs-v2/pull/100/files#diff-6614b363b750dec52fa4be5525d63c2cR46) in this edit then? I was assuming image paths work the same way in VuePress as in other static-site generators.

Currently, the images are sitting within an `images` folder in the same directory as the page itself.

Screen Shot 2020-02-12 at 4 59 43 PM

Is this the right place, or should they be somewhere else?

jessicaschilling:
@johnnymatthews See https://github.com/ipfs/ipfs-docs-v2/pull/110 for relative paths that work in both ... h/t @cwaring 😊

cwaring:
both examples are valid and should work, but it the first one causes any trouble then you now have an alternative 👍

cwaring:
relative paths keeps things IPFS and offline friendly so only use absolutes when it's absolutely necessary. this feels like another candidate for a linter rule 😅

PR #110, opened 2020-02-12
fix: raw -> relative

cwaring:


jessicaschilling:
@cwaring Thanks for finding the proper way of relative-linking these!
cc @johnnymatthews, for including future images.

cwaring:
Looks like this works @jessicaschilling ✅
https://github.com/ipfs/ipfs-docs-v2/blob/fix/relative-images/docs/community/contribute/grammar-formatting-and-style.md

With the added benefit of hashed images in production so that your cache doesn't serve us stale content:
CleanShot 2020-02-12 at 22 02 27

PR #112, opened 2020-02-13
Fix `mulit' typo in reference to `multihash'

qzdl:
Super trivial


jessicaschilling:
Thanks for finding and fixing!

PR #115, opened 2020-02-18
libp2p docs

edgo914:


jessicaschilling:
Thanks! This is a great start. What about expanding upon the bullet points for features a bit - basically just giving each of them a little bit more information, so the reader can go in depth a bit more without having to click through to the subsidiary pages?

It would also be great to see some of the reasons why someone would want to use libp2p - the sorts of ways it is useful, the problems it solves, etc. You can see some of this in the libp2p docs themselves: https://docs.libp2p.io/introduction/what-is-libp2p/

I'll leave the rest to @johnnymatthews to discuss, just chiming in 😊

johnnymatthews:
This PR is a great start, thanks a lot @edgo914. Like @jessicaschilling said, it'd be handy if you could expand on the bullet-points you've listed. There're a few formatting changes that need adding, but I'll wait until you've finished the rest of the article instead of throwing them in now.

johnnymatthews:
This PR can be closed. Everything here is addressed in [another PR](https://github.com/ipfs/ipfs-docs-v2/pull/122/files)

PR #153, opened 2020-04-01
fix typo in bitswap.md

davidB:


jessicaschilling:
Thanks for finding that! 😊

PR #164, opened 2020-04-09
HTTP Reference: update docs with new improvements from http-api-docs generator

hsanjuan:


hsanjuan:
Ugh this markdown parser is keeping line breaks and making them `<br />` :((

PR #166, opened 2020-04-09
http reference: Fix more char escaping issues

hsanjuan:
I am so sorry


jessicaschilling:
hahahaha, no worries! 😊 Thanks for the attention to detail.

PR #170, opened 2020-04-12
Fix a mistake in docs/concepts/dht.md

vojta001:
According to line 36, m is the number of bits, not the number of
combinations.


johnnymatthews:
Thanks again @vojta001!

PR #193, opened 2020-04-29
Fix links for js-ipfs repository

sinkuu:


johnnymatthews:
Thanks for sorting those out @sinkuu!

PR #195, opened 2020-05-01
export CLUSTER_HOST variable

mburns:
Not sure this works, still need to test

This is meant to trigger pinning the new website content to the Collab Cluster, which will fix the issues we've been having.


johnnymatthews:
Can I merge this in @mburns, or are you still testing things?

mburns:
talked this over at the Bifrost standup this morning, and settled on a different solution (https://github.com/protocol/bifrost-infra/pull/725). 

@johnnymatthews you can resume your regularly scheduled merge/deploys. Changes to the site should not cause errors anymore.

cwaring:
@mburns should we push to cluster for the new protocol.ai config too?

mburns:
@cwaring I believe so. Want to test that this works as expected first.

mburns:
I don't have Circle permissions to test this further, but it appears to be what we need.

PR #197, opened 2020-05-03
fix spelling

apuchitnis:


johnnymatthews:
Thanks for catching that one @apuchitnis!

PR #213, opened 2020-05-12
Correct YouTube overview links in concepts/README.md

EricLScace:
These should be corrected now.

ref https://github.com/ipfs/docs/issues/526

### DIFF
--- a/docs/concepts/README.md
+++ b/docs/concepts/README.md
@@ -12,9 +12,9 @@ New to IPFS or the distributed web? Here's where to get started on understanding
 
 Want the basics, in theater mode? Check out these Core Courses from IPFS Camp 2019 for an overview of how IPFS works with files and the overall data lifecycle in dweb.
 
-@[youtube](Z5zNPwMDYGg)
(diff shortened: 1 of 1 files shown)


johnnymatthews:
Nice, thanks for fixing that @EricLScace! Closes ipfs/docs#526.

cwaring:
@EricLScace @johnnymatthews just a heads up these were formatted this way because they are YouTube embeds on the website view of the application, if you want to also provide links to these items then you should add an inline link below the embed (however you could also simply click on the YouTube player from the docs site to view original video).

Changing these to links has removed the embedded player from the website:
2020-05-13 at 17 16 52

@EricLScace was this a problem because you were browsing the content on GitHub instead of at https://docs-beta.ipfs.io/concepts/?

EricLScace:
Yep, I was browsing on GitHub. I'll make a new PR…

johnnymatthews:
Ah, didn't know. Thanks for the heads up @cwaring.

PR #237, opened 2020-05-22
feat: new callout, update implementation status and tweaks

hugomrdias:
feat: add callout container
fix: links and remove issue from features
feat: complete implementation-status overhaul

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -370,28 +370,19 @@ module.exports = {
     }
   },
   plugins: [
-    ['@vuepress/plugin-back-to-top', true],
+    '@vuepress/back-to-top',
     [
-      '@vuepress/active-header-links',
+      '@vuepress/google-analytics',
       {
-        sidebarLinkSelector: '.sidebar-link',
-        headerAnchorSelector: '.header-anchor',
-        headerTopOffset: 120
+        ga: 'UA-96910779-15'
       }
     ],
-    '@vuepress/plugin-last-updated',
     [
       'vuepress-plugin-clean-urls',
       {
         normalSuffix: '/',
         indexSuffix: '/',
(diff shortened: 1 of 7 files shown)


docs/.vuepress/config.js
```diff
@@ -439,17 +430,22 @@ module.exports = {
...
       {
         host: DEPLOY_DOMAIN
       }
+    ],
+    [
+      'vuepress-plugin-container',
+      {
+        type: 'callout',
+        defaultTitle: ''
+      }
```
cwaring, line 448: this plugin is enabled in the parent theme so we shouldn't require it here too?

hugomrdias: to create a new container we need to add another array here

hugomrdias: https://github.com/vuepress/vuepress-plugin-container#usage

cwaring: sorry, what I was meaning is could we reuse one of the existing blocks? https://vuepress.vuejs.org/guide/markdown.html#custom-containers

cwaring: just noticed they print the type too so wouldn't be suitable. ok - let's go this route!

docs/.vuepress/config.js
```diff
@@ -370,16 +370,13 @@ module.exports = {
     }
   },
   plugins: [
-    ['@vuepress/plugin-back-to-top', true],
+    '@vuepress/back-to-top',
```
cwaring, line 373:
```suggestion
    '@vuepress/plugin-back-to-top',
```

`@vuepress/back-to-top` doesn't exist

cwaring:
This is good but please let's keep any further style and functionality changes out of this effort until we get to live. Time is limited atm any this work is extra to review.

PR #329, opened 2020-06-04
fix: new docsearch index

cwaring:
Algolia just migrated and deleted our old index, this connects to the new instance.


johnnymatthews:
Did Algolia email you? And can I get in on that email thread? Just in case you're away/not reachable if this happens again.

cwaring:
I've been in touch with them to update [our config](https://github.com/algolia/docsearch-configs/blob/master/configs/ipfs-docs.json). I've also requested to be connected to the new analytics dashboard.

PR #351, opened 2020-06-17
chore: update api reference

Stebalien:


johnnymatthews:
It's mildly annoying that the script uses `&#39;` instead of `'`, but it doesn't break anything I guess.

PR #406, opened 2020-08-14
Update ipfs-gateway.md

momack2:
Added an image of all the different public gateways


johnnymatthews:
Good idea. I'm gonna crop the image to hide the browser, and throw it into the repo so it gets _IPFSified_.

PR #407, opened 2020-08-15
Do not claim that IPNS addresses are human-readable

carlbordum:
Before this commit, the docs claim that IPNS solves two problems: stable and human-readable names, but in fact in only solves the former. Currently the solution for human-readable names is DNSLINK.


johnnymatthews:
That's a fair point. Thanks for cleaning that up for us.

PR #418, opened 2020-08-23
Fix Fleek link

cqcallaw:
The single-dot link resolves to https://docs.ipfs.io/how-to/websites-on-ipfs/link-a-domain/introducing-fleek instead of the desired https://docs.ipfs.io/how-to/websites-on-ipfs/introducing-fleek/


johnnymatthews:
Ah nice! Thanks for the catch!

PR #470, opened 2020-09-24
remove docs-beta dnslink entry

thattommyhall:
docs-beta is redirected to docs now, do you still need the old dnslink?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,6 @@ jobs:
       - image: olizilla/ipfs-dns-deploy:debug
         environment:
           DOMAIN: docs.ipfs.io
-          DOMAIN_BETA: docs-beta.ipfs.io
           BUILD_DIR: ./docs/.vuepress/dist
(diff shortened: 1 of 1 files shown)


thattommyhall:
Could also set _dnslink of docs-beta to `/ipns/docs.ipfs.io` if it is to always be same?

johnnymatthews:
Nope, I don't think we do. Should probably check with @cwaring though, just to make sure.

lidel (commented):
If we remove this, then we also should remove the DNS TXT record, otherwise people with companion will see some old version. 

See: https://github.com/ipfs/ipfs-docs/pull/367

johnnymatthews:
@cwaring can we merge this?

cwaring:
@johnnymatthews sure with the manual step to remove the stale TXT that @lidel suggested. @thattommyhall could you take that one?

thattommyhall:
dnslink for docs-beta is deleted

cwaring:
Perfect, I'll merge now so that it doesn't get recreated.

Thanks all!

PR #471, opened 2020-09-24
Add other persistence options to how to pin files

momack2:
Valid callout from users that pointing to pinata as the only option for pinning isn't accurate and creates a centralization vector. Let's also highlight the other options. This is a really fast draft, would love review/refinement from @johnnymatthews 🙏


johnnymatthews:
Cleaned things up a little bit. The only think we're missing is a list of IPFS pinning services. Adding this comment as a placeholder for that info:

- [Pinata](https://pinata.cloud/)
- [Temporal](https://temporal.cloud/)

PR #481, opened 2020-09-30
Ubuntu via snap and .deb install

NukeManDan:
Options added, clarify where to get .deb

### DIFF
--- a/docs/install/ipfs-desktop.md
+++ b/docs/install/ipfs-desktop.md
@@ -74,7 +74,17 @@ The IPFS Desktop application has finished installing. You can now start to [add 
 
 While these instructions are specific to Ubuntu, they will likely work with most Ubuntu-related Linux distributions. For non-Ubuntu Linux distributions, check out the [IPFS Desktop GitHub repository](https://github.com/ipfs-shipyard/ipfs-desktop#install) for install instructions.
 
-1. Download the `.deb` package:
+### Install Option 1: Snap Package
+
+If you with to use the [`snap` package manager](https://snapcraft.io/) - all maintained Ubuntu versions have this by default - open up a terminal and install via:
+
+```bash
+$ snap install ipfs-desktop
+```
+
+### Install Option 2: `.deb` Package
+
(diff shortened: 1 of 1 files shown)


johnnymatthews:
Oh wow! Don't know how we missed this one. Thanks for bringing it up @NukeManDan! I like your idea of where users can get the Snap package, but I'm not a huge fan of splitting up the **Ubuntu** section into **Option 1** and **Option 2**. It'd be handy to give users a quick reference to install IPFS Desktop with a package manager though. Maybe that could come at the top of this page under a **Package manager** section? But I'll leave that for a different PR/issue, since it's not a priority right now.

NukeManDan:
Sounds good @johnnymatthews - Would it be good to add an issue to track this so we can add it latter on? `snap` is easiest by far to install, but comes with caveats. Appimage IMHO is next in line, and lastly .deb install. 

Maybe this could be easiest to put in the docs as expandable sections? This is easy in markdown too: https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab

bertrandfalguiere:
One of the caveat if snap is that the doles are in a different place than with .deb. if you install one you won't see the files added with the other.

johnnymatthews:
Just made #486 to address this @NukeManDan. The docs site is built on VuePress, which allows for [expandable sections using ` ```details`](https://vuepress.vuejs.org/guide/markdown.html#custom-containers), so we could use those.

@bertrandfalguiere not sure what you mean by _doles_ (a typo?).

bertrandfalguiere:
I meant files. Damn autocorrect.

johnnymatthews:
That makes sense. Yeah I guess a user _could_ install IPFS Desktop via `apt` and `snap`. We should mention that this is generally a bad idea.

bertrandfalguiere:
ipfs-desktop and go-ipfs are not published to `apt`, unfortunately.
image


But they can install via a downloaded `deb` package and then via `snap`. I got bitten myself: it is confusing when you are new to IPFS.

johnnymatthews:
Sorry, should've been clearer. I meant using `apt` to install a local `.deb` file:

```bash
sudo apt install ./ipfs-desktop.deb
```

PR #488, opened 2020-10-06
Add an explicit cd command to copypaste

mburns:
This matches the equivalent instructions for Linux users later in the file (L145).

I just want to be able to blindly copy instruction and Have It Work(TM).


johnnymatthews:
Ah, nice catch! It might be useful to explore adding a _one-liner_ to each of the OSes where possible, so folks can just copy-and-paste a single command into the terminal. I'll make an issue for it.

Anyway, thanks for the PR!

PR #496, opened 2020-10-12
Fix VuePress getting started link

cqcallaw:
Previously the link pointed to Hugo's Quick Start


johnnymatthews:
Nice, thanks for the assist!

PR #497, opened 2020-10-13
Fix: Remove hashes from protoschool links (possible cause of mislabeled source in ProtoSchool Countyl stats)

zebateira:
This is an optimisation to reduce the number of redirects ProtoSchool users need to go through, but also a possible fix for the lower source traffic in ProtoSchool from docs.ipfs.io, which we will monitor over the following weeks.
After some time we will do another PR to add a main link in the homepage to ProtoSchool like @jessicaschilling suggested.


johnnymatthews (approved):

zebateira:
thank you @johnnymatthews 🙏

Issue #480, opened 2020-09-28
Evaluate the "Recent releases" section.


johnnymatthews:
The [Recent releases](https://docs.ipfs.io/recent-releases/) section was created for the Go-IPFS 0.5 release to showcase the incredible amount of features, fixes, and optimizations that went into that release. Since then, every major release of Go-IPFS has been listed in that section. However, most of the information contained within the section is copied from blog posts, changelogs, and release notes created within the `ipfs/go-ipfs` repo. This section isn't adding any new information and only serves to advertise which versions are available, along with a very summarised feature list.

Should we keep this section around? If yes, it should be refactored so that it's smaller, easier to read, and doesn't just copy-and-paste from [blog.ipfs.io](https://blog.ipfs.io/).

johnnymatthews:
At the _very least_ we can change how things show up in the sidebar by collapsing **GO-IPFS 0.7**, **GO-IPFS 0.6**, etc:

![image](https://user-images.githubusercontent.com/9611008/94486061-413a9a00-01b5-11eb-8a1f-72d8c02cf308.png)

PR #511, opened 2020-10-17
Collapse Latest Release in sidebar

ChaituVR:
Related to #480 

image


johnnymatthews:
Nice! This doesn't actually _close_ #480 since we still need to decide whether or not to keep that section around. But having the releases collapsible does answer the second comment in that issue. So thanks!

PR #541, opened 2020-11-16
Update ipfs-updater.md

yuzibo:
Darwin is macos, now it is linux binary package for ipfs-update


johnnymatthews:
Ah, looks like that was a copy-and-paste fail. Thanks for fixing!

PR #543, opened 2020-11-20
Update generate-cli-docs.sh

yuzibo:
Maybe small typo fix.


johnnymatthews:
Nice catch! Thanks @yuzibo.

PR #547, opened 2020-11-24
Improves readability

pr0PM:
Just got confused for a moment what `vidhash` was, so why not. Might change to something better if addressed/suggested.


johnnymatthews:
Thanks for pointing this out! Simplified things a bit more.

PR #555, opened 2020-12-04
Adds to Getting started section.

### LINKED ISSUE #553: Create high-level map for the "Getting started" section.
# Getting Started

To create a simple flow for new readers coming to the IPFS docs, we need to have a really good Getting started section that covers the main points of IPFS:

- What is IPFS
- Why IPFS exists, and the current problems with Web 2.0
- How does IPFS solve these problems

## What: a simple definition of IPFS.

- Set of protocols that allow computers to share data.
- Based around peer-to-peer networks, instead of relying on central servers.
- Designed as a replacement for HTTP that can grow to an interplanetary scale. 

## Why: current problems with Web 2.0.

- Location bases addresses are bad:
	- Users have to know exactly where the data is.
	- If the data moves, they need a new address.
	- There's no in-built way to check that the data you get back is correct.
	- With current infrastructure, there's a much bigger risk of data becoming unavailable (blocked, lost, or modified).

### Content addressing

Content-addressing makes it easier to get specific file versions, create multiple copies of stuff, and allow anyone in the world to send you data. even people you don't trust) since it automatically checks that the data you're getting is the data you asked for in the first place.

This means I can fetch data from someone nearby (e.g., on my planet) instead of going to some specific location (e.g., Earth). Basically, centralized systems start to fall apart as latencies increase.

### Bitswap

Since many people all over the world can be sending you data, we can make protocols that balance fetching data from many people at the same time. This is why Bitswap was created.

### Distributed hash-table

While content addressing means we don't need to know in advance who will send us data, since it could be anyone, we still need to find a way of collecting a list of people who might be able to give us that data. The IPFS Public DHT is one of the ways IPFS tackles this problem.

## Other points

It's interesting to think of the cost-benefit of both IPFS and the current web 2.0 protocols. The benefit of web 2.0 protocols is that it's easy to get the data that you're looking for, but it comes at the cost of needing to trust the server you're getting it from. IPFS benefits from efficiently fetching data from many places at once, preferring close peers and inexpensive connections. But that comes at the cost of having to find who has the data in the first place. If you happen to know already then great! Otherwise we have some schemes that let you advertise that you have data and help other people find out you have it.

johnnymatthews:
Create a single, obvious starting point for a developer to begin getting familiar with IPFS. The current introduction guide focuses on the IPFS CLI, and assumed the reader to already understand concepts like CIDs and pinning.

Much of the content already exists, so this project is just a simple case of getting everything in one place, and reorganizing it.

Closes #553


docs/getting-started/README.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Getting started
+description: Start your IPFS journey here! Learn how the Interplanitery File System works, install commonly used tools, and get to grips with basic d-web concepts.
+---
+
+The InterPlanetary File System (IPFS) is a set of protocols that allow computers to share data using peer-to-peer networks instead of relying on central servers. The project was started as a response to major issues with the internet, and is designed as a replacement for HTTP that can grow to an interplanetary scale. This section covers the issues that IPFS fixes, how it manages those solutions, and how you can build on top of the protocol.
```
AlexxNica, line 6:
```suggestion
The InterPlanetary File System (IPFS) is a set of protocols that allow computers to share data using peer-to-peer networks instead of relying on central servers. The project started as a response to critical issues we currently face on the internet and is designed to replace HTTP and grow to an interplanetary scale. This section covers the issues that IPFS fixes, how it manages those solutions, and how you can build on top of the protocol.
```

Minor improvements to readability.

docs/getting-started/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Problems with the current internet
+
+Currently, the computers on the internet share files using fixed-addresses. If Laika wants to get a file on Albert's computer, Laika needs to know:
+
+1. The IP address of your Albert's computer.
+2. The location of the file.
+3. The filename itself.
+
+This comes with problems. What if Albert's computer is offline or unavailable? Albert _could_ store a copy of the file on another computer for redundancy, but then Laika also needs to know the address of _that_ computer. That backup computer would also need the exact same directory structure as Albert's original computer. What if Albert moves the file to a different location on their computer? Now the address Laika has is wrong. And if Laika manages to find the new address, the address of the file on the backup computer is still wrong, so you'd have to update the location of the file on _that_ computer as well.
```
AlexxNica, line 16:
```suggestion
This approach comes with problems. What if Albert's computer is offline or unavailable? Albert _could_ store a copy of the file on another computer for redundancy, but then Laika also needs to know the address of _that_ computer. That backup computer would also need the exact same directory structure as Albert's original computer. What if Albert moves the file to a different location on their computer? Now the address Laika has is wrong. And if Laika manages to find the new address, the file's address on the backup computer is still incorrect, so you'd have to update the file's location on _that_ computer as well.
```

Minor improvements to readability.

AlexxNica (commented):
I've added some comments that I think can help improve this PR! I hope it helps. 😁

docs/getting-started/next-up.md
```diff
@@ -0,0 +1,4 @@
+---
+title: Next up
+description: Start your IPFS journey here! Learn how the Interplanitery File System works, install commonly used tools, and get to grips with basic d-web concepts.
```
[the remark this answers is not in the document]
johnnymatthews: Oh wow, thanks for finding this info @AlexxNica!

> We can also create an issue to discuss this.

I don't really want to open this up in an issue for fear of bike-shedding, but we can do if folks think a discussion would be beneficial. 

I reckon we should change any instances of _DWeb_, _dweb_, and _dWeb_ instances to _d-web_, since these read like the word _dweeb_. Separating the _d_ from the _web_ encourages readers to internally pronounce the two syllables separately.

johnnymatthews:
Closing until I devote more time to this.

PR #559, opened 2020-12-10
Fixed code block escaping

eabadjiev:


johnnymatthews:
There seems to be a few of these floating around for some reason. Thanks for catching this!

PR #571, opened 2020-12-16
docs: inline opened w/ single quote, should be backtick

plantimals:
Replaced the single quote with a backtick and the formatting is restored


johnnymatthews:
Nice find! Looks like that's the only one in there too. Cheers!

image

PR #575, opened 2021-01-02
Fix broken link

Sosthene-Guedon:
The link to the Fleek case study is broken. If you go on [the docs](https://docs.ipfs.io/concepts/privacy/#encrypting-content-transported-via-ipfs) and click on the link `this case study on Fleek, a fast-growing IPFS file hosting and delivery service`, you will get a 404 error. This fixes it.

tmp_ipfs_docs


johnnymatthews:
Nice catch. Thanks for the PR!

PR #592, opened 2021-01-09
Use promise instead of a callback as argument in ipfs.name.publish example

marc2332:
## Description
In October 21th the method `ipfs.name.publish` removed the ability to pass a function as callback. Currently, you can only use the promise-based callback. This PR updates the docs, so the example uses a promised-based callback.

Link to the docs page: https://docs.ipfs.io/concepts/ipns/#example-ipns-setup-with-js-sdk-api

Old (callback as argument):

image

## Context
This is the commit that removed the function as argument callback: https://github.com/ipfs/js-ipfs/commit/bbcaf34111251b142273a5675f4754ff68bd9fa0#diff-9899aad8c28df56d9808f76d9f98c23a49ec9cd4eadeeb888c1e1098ac17b245

image

## Result

My fix (promised-based callback):

image

### DIFF
--- a/docs/concepts/ipns.md
+++ b/docs/concepts/ipns.md
@@ -119,7 +119,7 @@ Here's where the Name API comes in handy. With it, you can create a single, stab
 // The address of your files.
 const addr = '/ipfs/QmbezGequPwcsWo8UL4wDF6a8hYwM1hmbzYv2mnKkEWaUp'
 
-ipfs.name.publish(addr, function (err, res) {
+ipfs.name.publish(addr).then(function (res) {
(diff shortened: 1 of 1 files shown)


johnnymatthews:
Oh wow, nice catch @marc2332. And thanks for summarizing your PR so well! It really helps when contributors show their process and reasons for making the PR, so thanks for doing that :1st_place_medal:

johnnymatthews:
Recreated the PR in #596 since it looks like CircleCI is stuck here. Closing this one. Thanks for your help @marc2332!

PR #598, opened 2021-01-14
Only pin master builds to collaborative cluster

hsanjuan:
Every website build is sent to the collaborative cluster, but it should rather put those in the legacy cluster and only add the master builds into the collaborative one.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
             
             hash=$(pin-to-cluster.sh "$pin_name" /tmp/workspace/$BUILD_DIR)
             echo "Website added to IPFS: https://ipfs.io/ipfs/$hash"
-            # Update DNSlink for prod or dev domain
+            # Update DNSlink for prod
             if [ "$CIRCLE_BRANCH" == "main" ] ; then
               dnslink-dnsimple -d $DOMAIN -r _dnslink -l /ipfs/$hash
               export CLUSTER_HOST=/dnsaddr/ipfs-websites.collab.ipfscluster.io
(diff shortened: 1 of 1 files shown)


hsanjuan:
@cwaring @johnnymatthews this should be similar for any other websites you manage.

cwaring:
Thanks @hsanjuan!

This doesn't look quite right, we already have the `CLUSTER_HOST` [env set](https://github.com/ipfs/ipfs-docs/blob/main/.circleci/config.yml#L36) and have also recently changed `master` to `main`.

The pinning happens on [this line](https://github.com/ipfs/ipfs-docs/blob/main/.circleci/config.yml#L44), which we can move into the if statement but then how long will unpinned build previews remain on the cluster for? It's convenient to use these preview links for review occasionally.

hsanjuan:
@cwaring you're right. For such a small PR it had like 20 errors. See now?

Note everything's safely pinned, just not everything in the collaborative cluster. The default is the large storage cluster.

.circleci/config.yml
```diff
@@ -46,6 +45,9 @@ jobs:
             # Update DNSlink
             if [ "$CIRCLE_BRANCH" == "main" ] ; then
               dnslink-dnsimple -d $DOMAIN -r _dnslink -l /ipfs/$hash
+              export CLUSTER_HOST=/dnsaddr/ipfs-websites.collab.ipfscluster.io
+              pin-to-cluster.sh "$DOMAIN" /tmp/workspace/$BUILD_DIR
+              echo "Website added to collaborative cluster"
```
cwaring, line 50: This will break the live dnslink because `$hash` is no longer defined. What was wrong with the previous script line and using docker to set `CLUSTER_HOST`?

cwaring: ah yes, I missed that in the folded code view. Now I see what you're doing here :)

cwaring (approved):

cwaring:
Thanks, I will merge this now and if things work I can roll out to other sites too!

PR #605, opened 2021-01-22
Add a note about Snap confinement

gammazero:
Users who want to know how to install IPFS may not be aware of Snap's [confinement](https://snapcraft.io/docs/snap-confinement) behavior.  So, I thought it useful to add a note someplace where they might see it before they install the Ubuntu Snap package.

Hopefully, this can prevent issues like this one: https://github.com/ipfs/go-ipfs/issues/7872

### DIFF
--- a/docs/install/command-line.md
+++ b/docs/install/command-line.md
@@ -22,7 +22,7 @@ Most operating systems have a package manager, although they may not come pre-in
 Although we try our best to keep the package manager releases up to date, they sometimes lag behind the Go-IPFS GitHub releases page by a few days. If you'd like to install a release the very same day it comes out, use the [official distributions](#official-distributions) from [dist.ipfs.io](https://dist.ipfs.io).
 
 :::warning
-On Ubuntu, the IPFS Snap uses [strict confinement](https://snapcraft.io/docs/snap-confinement) by default and might affect usability of IPFS. Consider specifying the `--classic` option when installing.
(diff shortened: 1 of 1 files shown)


aschmahmann:
cc @olizilla, does this sound good to you?

olizilla (approved):
Yes indeed! Good idea to warn about it. I would use a `::: callout` or similar to highlight it, and move it to directly below the table.

johnnymatthews:
Moved the warning into a `:::warning` box. Saying that, if installing things through Snap _without_ the `--classic` flag is gonna cause issues, should we just advise folks to use the `--classic` flag all the time?

aschmahmann:
@gammazero I'm not sure this is actually enough. Since the snap has `strict` in the config.yaml users get `Warning: flag --classic ignored for strictly confined snap ipfs`.

We should probably just change to telling people to use `--classic` (and changing our config.yaml). We could potentially give access to more plugs as well (perhaps, `system-files`).

johnnymatthews:
Updated the instructions to use `--classic` by default.

> changing our config.yaml - @aschmahmann 

What's required for this?

olizilla:
The snap default is strict confinement. The ipfs snap has always had strict confinement. Are lots of folks getting suprised by that? The installing user can add and get files via an ipfs daemon installed by snap, so I'm not clear on what the problem is.

Snap is not other package managers, and we should be careful around our assumptions of how things should work. We should definitely review the list plugs and add any that are missing tho. I am mildly against reducing the confinment at this point unless i am missing some context.

johnnymatthews:
Right. In that case, I'm not sure what to do here then. Installing IPFS with Snap is easy to set up and for most folks to use, but comes with confinement complications. I'd like to keep a simple one-liner here for Ubuntu users though.

olizilla:
The best thing would be to caveat it in the docs for now, similar to the orginal PR, and open an issue on go-ipfs to review the snap.yml config to make it as un-annoying as possible, given that snaps defaults are much stricter than other package managers.

johnnymatthews (approved):

PR #616, opened 2021-01-28
Changed docsBranch value to main

LuisEnMarroquin:
`Edit this page` href currently opens url with `master` branch but is has been changed to `main`

image


cwaring:
Thanks for the contribution @LuisEnMarroquin. I have now fixed this upstream in the main app config 👍🏻

PR #623, opened 2021-01-30
Fix italics formatting

Dietr1ch:
_ things with a leading space are not parsed correctly_


johnnymatthews:
Looks like a find-and-replace gone wrong. Thanks for catching this one. We merge to `main` on Tuesday and Thursdays, so this should get merged in then.

PR #626, opened 2021-02-02
Mention browsers that support IPFS.

jurjendevries:
Now some browsers have IPFS support built-in it would be great to mention them.


johnnymatthews:
Excellent point! I've added links to each browser's website.

johnnymatthews (approved):

PR #628, opened 2021-02-05
Update command-line.md

fte378:
Add OpenBSD package to list of operating systems and package managers


johnnymatthews:
Thanks for this PR @fte378! There are a couple of issues with it however:

This section needs to be kept short, and OpenBSD isn't popular enough to fit in here. Also, if we add an OpenBSD one-liner here, then we'd also have to include OpenBSD instructions throughout the docs to keep consistency. OpenBSD does not have a big enough market-share to warrant that.

It's safe to assume that OpenBSD users are advanced enough to find the `pkg_add go-ipfs` package and install it themselves.

A case could be made for creating an _Alternative OS installation_ section where we list the standard ways of installing IPFS on various operating systems. I'd encourage you, or anyone else reading this, to create an issue for this, if you think it's relevant enough. As it stands right now, I'm going to close this PR.

PR #656, opened 2021-02-21
Explain why hashes are different between systems.

### LINKED ISSUE #430: Explain why hashes are different between systems.
# [This issue has a bounty of $200 USD](https://github.com/ipfs/devgrants/projects/1)

This issue cropped up from a discussion in IPFS IRC.

It somewhat frequently comes up that people ask or wonder why if they do `ipfs add ubuntu.iso` the resulting SHA256 _multihash_ has nothing to do with what they'd get running `sha256sum ubuntu.iso` (or with the hash posted on the Ubuntu website). Depending on the user, this question can come across using various levels of our jargon (e.g., I'm looking for the hash, not the multihash, or how do I turn a CID into a hash), but the idea is the same.

It may be useful to explain why this happens and what's going on. This issue is also likely a UX problem with IPFS itself, and might not be fixable from a purely docs perspective.

Here's an explanation put forward by @aschmahmann:

> CIDs are a format for used for identified content in a self describing way. They come with a base encoding, codec (i.e. what kind of data is the CID referring to), and a multihash.
> 
> Multihashes are a format for identifying hashes in a self describing way. Their point is to make it obvious that the thing you are looking at is a SHA-512 that's been truncated to 256 bits instead of a SHA-256 in it's entirety (it turns out people actually do these sorts of things with hash functions).
>
> IPLD utilizes the properties of merkle trees where if I have two pieces of data with hashes A and B, then I can refer to both of those pieces of data together uniquely by hashing the data structure that contains A and B and gives me C. As a result for any arbitrarily large tree of data identified by hashes there is a root node R whose hash identifies the entire tree including all of its children.
> 
> UnixFS is a particular IPLD format that was designed to take files, folders, etc. and structure them as merkle trees. For example, it's very helpful to p2p data transfer protocols like bitswap that instead of asking for a single 100 MB file they can ask for 100 one MB files. The root of this merkle tree that has many nodes that ultimately contain a 100MB file identifies the entire 100MB file as well as the graph structure/layout of the small blocks of data that make up the tree.
> 
> When you say that you have the CID of a file what you're really saying is that you have a reference to the root of a tree that includes all the data in your file.
> 
> This means that your question is ultimately: "Is there a way for me to determine the hash of the "file" part of the data given only the hash of the root of the tree that contains the file data?".
>
> In general the answer is no. For example, while it's obvious how to sha256 a file how do you sha256 a directory, as there are multiple ways the data could be structured. In the particular case of getting the sha256 of UnixFS files the answer is also no. However, it's certainly possible to create combinations of hash functions and data structures such that HashOfRoot = MyHashFunction(FileData) (as a dumb example MyHashFunction(FileData) could be just taking the result of ipfs add FileData). Is there a way to create a data structure for storing a file in a tree where the root hash happens to give you sha256 of the file data? Not that I know of, but perhaps possible.
>
> As an aside: CIDs are pretty generally useful. For example, if you have the SHA256 of some file (e.g. an Ubuntu ISO) you might decide to describe that using a CID (e.g. with a raw codec) so that it's easier to work with and becomes self describing. 

## Tasks

- [ ] Add a new section to [/concepts/hashing/](https://docs.ipfs.io/concepts/hashing/#hashes-are-important) called something like _Not all hashes are the same_ or something along those lines.
- [ ] Bring up the example above, stating that _the hashes Ubuntu shows for their `ISO`s are different from IPFS hashes because_...
- [ ] Show the math of how hashes are calculated using an incredibly simple hashing algorithm.
    - [ ] Change one variable in that algorithm, but feed it the same input string to show how different hashing functions produce different hashes, even when hashing the same file.

alexmmueller:
Hi @johnnymatthews,

this is my first draft for  [#430]. I decided to not add another hint on how input changes change the output of a has function as this is IMO not really related to the task at hand and already captured quite well in the existing documentation. Can you let me know if this generally goes into the right direction and what you would like to see improved / added? Thank you! 

Closes #430.

### DIFF
--- a/docs/concepts/hashing.md
+++ b/docs/concepts/hashing.md
@@ -101,7 +101,7 @@ ubuntu-20.04.1-desktop-amd64.iso: FAILED
 shasum: WARNING: 1 computed checksum did NOT match
 ```
 
-As we can see the hash included in the CID does NOT match the hash of the input file `ubuntu-20.04.1-desktop-amd64.iso`. To understand what the hash included in the CID actually is, we have to understand how IPFS stores files. IPFS uses a so called Distributed Acyclic Graph (DAG) to keep track of all the data stored in IPFS. A CID identifies one specific node in this graph. This identifier is the result of hashing the node's contents — any opaque payload carried by the node and the list of identifiers of its children — using a cryptographic hash function like `SHA256`. To learn more about wow the DAG is used within IPFS to address content is explained in detail in the article covering the [Merkle DAG](/concepts/merkle-dag/).
+As we can see, the hash included in the CID does NOT match the hash of the input file `ubuntu-20.04.1-desktop-amd64.iso`. To understand what the hash contained in the CID is, we must understand how IPFS stores files. IPFS uses a [distributed acyclic graph (DAG)](/concepts/merkle-dag/) to keep track of all the data stored in IPFS. A CID identifies one specific node in this graph. This identifier is the result of hashing the node's contents using a cryptographic hash function like `SHA256`.
 
 ::: tip
 The [Tutorial: Merkle Trees and the IPFS DAG](https://dweb-primer.ipfs.io/ipfs-dag) inlcudes in-depth lessons coveing the IPFS DAG.
--- a/docs/concepts/hashing.md
+++ b/docs/concepts/hashing.md
@@ -101,4 +101,8 @@ shasum: WARNING: 1 computed checksum did NOT match
 The names of hash functions are not used consistently.`SHA-2`, `SHA-256` or `SHA-256 bit` all refer to the same hash function.
 ::: tip
 
-As we can see the hash included in the CID does NOT match the hash of the input file `ubuntu-20.04.1-desktop-amd64.iso`. To understand what the hash included in the CID actually is, we have to understand how IPFS stores files. IPFS stores files as a set of data chunks. A data structure called the DAG - or 'Directed Acyclic Graph' keeps track of which data chunks actually make up a certain file. The hash included in the CID identifies a node within the DAG. From this node all data chunks that represent our input file are linked and can thus be retrieved by IPFS.
+As we can see the hash included in the CID does NOT match the hash of the input file `ubuntu-20.04.1-desktop-amd64.iso`. To understand what the hash included in the CID actually is, we have to understand how IPFS stores files. IPFS uses a so called Distributed Acyclic Graph (DAG) to keep track of all the data stored in IPFS. A CID identifies one specific node in this graph. This identifier is the result of hashing the node's contents — any opaque payload carried by the node and the list of identifiers of its children — using a cryptographic hash function like SHA256. To learn more about wow the DAG is used within IPFS to address content is explained in detail in the article covering the [Merkle DAG](/concepts/merkle-dag/).
+
+::: tip
+The [Tutorial: Merkle Trees and the IPFS DAG](https://dweb-primer.ipfs.io/ipfs-dag) includes detailed lessons on how to work with the IPFS DAG.
+::: tip
--- a/docs/concepts/hashing.md
+++ b/docs/concepts/hashing.md
@@ -101,8 +101,8 @@ shasum: WARNING: 1 computed checksum did NOT match
 The names of hash functions are not used consistently.`SHA-2`, `SHA-256` or `SHA-256 bit` all refer to the same hash function.
 ::: tip
 
-As we can see the hash included in the CID does NOT match the hash of the input file `ubuntu-20.04.1-desktop-amd64.iso`. To understand what the hash included in the CID actually is, we have to understand how IPFS stores files. IPFS uses a so called Distributed Acyclic Graph (DAG) to keep track of all the data stored in IPFS. A CID identifies one specific node in this graph. This identifier is the result of hashing the node's contents — any opaque payload carried by the node and the list of identifiers of its children — using a cryptographic hash function like SHA256. To learn more about wow the DAG is used within IPFS to address content is explained in detail in the article covering the [Merkle DAG](/concepts/merkle-dag/).
+As we can see the hash included in the CID does NOT match the hash of the input file `ubuntu-20.04.1-desktop-amd64.iso`. To understand what the hash included in the CID actually is, we have to understand how IPFS stores files. IPFS uses a so called Distributed Acyclic Graph (DAG) to keep track of all the data stored in IPFS. A CID identifies one specific node in this graph. This identifier is the result of hashing the node's contents — any opaque payload carried by the node and the list of identifiers of its children — using a cryptographic hash function like `SHA256`. To learn more about wow the DAG is used within IPFS to address content is explained in detail in the article covering the [Merkle DAG](/concepts/merkle-dag/).
 
 ::: tip
-The [Tutorial: Merkle Trees and the IPFS DAG](https://dweb-primer.ipfs.io/ipfs-dag) includes detailed lessons on how to work with the IPFS DAG.
+The [Tutorial: Merkle Trees and the IPFS DAG](https://dweb-primer.ipfs.io/ipfs-dag) inlcudes some in-depth lessonson the IPFS DAG and are a great ressource if you want to understand how IPFS uses the merkle .
 ::: tip
--- a/docs/concepts/hashing.md
+++ b/docs/concepts/hashing.md
@@ -49,7 +49,7 @@ Cryptographic hashes come with a couple of very important characteristics:
 - **unique** - it's infeasible to generate the same hash from two different messages
 - **one-way** - it's infeasible to guess or calculate the input message from its hash
 
-It turns out these features also mean we can use a cryptographic hash to identify any piece of data: the hash is unique to the data we calculated it from and it's not too long (a hash is a fixed length, so the SHA-256 hash of a one-gigabyte video file is still only 32 bytes), so sending it around the network doesn't take up a lot of resources.
+These features also mean we can use a cryptographic hash to identify any piece of data: the hash is unique to the data we calculated it from and it's not too long so sending it around the network doesn't take up a lot of resource. A hash is a fixed length, so the SHA-256 hash of a one-gigabyte video file is still only 32 bytes. 
 
 That's critical for a distributed system like IPFS, where we want to be able to store and retrieve data from many places. A computer running IPFS can ask all the peers it's connected to whether they have a file with a particular hash and, if one of them does, they send back the whole file. Without a short, unique identifier like a cryptographic hash, that wouldn't be possible. This technique is called [content addressing](/concepts/content-addressing/) — because the content itself is used to form an address, rather than information about the computer and disk location it's stored at.
 
--- a/docs/concepts/hashing.md
+++ b/docs/concepts/hashing.md
@@ -80,7 +80,7 @@ added QmPK1s3pNYLi9ERiq3BDxKa4XosgWwFRQUydHUtz4YgpqB ubuntu-20.04.1-desktop-amd6
  2.59 GiB / 2.59 GiB [==========================================================================================] 100.00%
 ```
 
-The string `QmPK1s3pNYLi9ERiq3BDxKa4XosgWwFRQUydHUtz4YgpqB` returned by the `ipfs add` command is the content identifier of the file `ubuntu-20.04.1-desktop-amd64.iso`. The content identifier contains various information in on human readable format. We can utilize the [CID Inspector](https://cid.ipfs.io/) in order to see what the CID includes. The actual hash is listed under `DIGEST (HEX)`:
+The string `QmPK1s3pNYLi9ERiq3BDxKa4XosgWwFRQUydHUtz4YgpqB` returned by the `ipfs add` command is the content identifier of the file `ubuntu-20.04.1-desktop-amd64.iso`. We can utilize the [CID Inspector](https://cid.ipfs.io/) to see what the CID includes. The actual hash is listed under `DIGEST (HEX)`:
 
 ```
 NAME: sha2-256
--- a/docs/concepts/hashing.md
+++ b/docs/concepts/hashing.md
@@ -57,15 +57,13 @@ That's critical for a distributed system like IPFS, where we want to be able to 
 
 It is important though to understand that a content identifier for a specific file added to IPFS does not equal the output of the respective hash function for that file. Hash functions are widely used as to check for file integrity. The downlooad provider published the output of a hash function for a file next to the file download. This output is called the checksum of that file. The checksum enables downloaders of the file to verify that the has not been altered since it was published or during the dowload. This check is done by performing the same hash function that was used to generate the checksum. If the local output matches the checksum, the file is verified and was not altered.
 
-Lets look at a concrete example: When you download the image file for the Ubuntu Linux distribution you might see the following SHA-256 checksum on the website:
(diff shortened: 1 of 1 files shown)


docs/concepts/hashing.md
```diff
@@ -30,7 +30,7 @@ However, the exact same input generates the following output using **SHA-256**:
 0x64EC88CA00B268E5BA1A35678A1B5316D212F4F366B2477232534A8AECA37F3C
 ```
 
-Notice that the second hash is longer than the first one. This is because SHA-1 creates a 160-bit hash, while SHA-256 creates a 256-bit hash. Also, the prepended `0x` is just an indicator that tells us that the following hash is represented as a base 16 (or hexadecimal) number.
+Notice that the second hash is longer than the first one. This is because SHA-1 creates a 160 bit hash, while SHA-256 creates a 256 bit hash. Also, the prepended `0x` is just an indicator that tells us that the following hash is represented as a base 16 (or hexadecimal) number.
```
johnnymatthews, line 33:
```suggestion
Notice that the second hash is longer than the first one. This is because SHA-1 creates a 160 bit hash, while SHA-256 creates a 256 bit hash. The prepended `0x` indicates that the following hash is represented as a hexadecimal number.
```

docs/concepts/hashing.md
```diff
@@ -49,6 +49,56 @@ Cryptographic hashes come with a couple of very important characteristics:
 - **unique** - it's infeasible to generate the same hash from two different messages
 - **one-way** - it's infeasible to guess or calculate the input message from its hash
 
-It turns out these features also mean we can use a cryptographic hash to identify any piece of data: the hash is unique to the data we calculated it from, and it's not too long (a hash is a fixed length, so the SHA-256 hash of a one-gigabyte video file is still only 32 bytes), so sending it around the network doesn't take up a lot of resources.
+It turns out these features also mean we can use a cryptographic hash to identify any piece of data: the hash is unique to the data we calculated it from and it’s not too long (a hash is a fixed length, so the SHA-256 hash of a one-gigabyte video file is still only 32 bytes), so sending it around the network doesn't take up a lot of resources.
```
johnnymatthews, line 52:
```suggestion
It turns out these features also mean we can use a cryptographic hash to identify any piece of data: the hash is unique to the data we calculated it from and it's not too long (a hash is a fixed length, so the SHA-256 hash of a one-gigabyte video file is still only 32 bytes), so sending it around the network doesn't take up a lot of resources.
```

docs/concepts/hashing.md
```diff
@@ -49,6 +49,56 @@ Cryptographic hashes come with a couple of very important characteristics:
...
 
 That's critical for a distributed system like IPFS, where we want to be able to store and retrieve data from many places. A computer running IPFS can ask all the peers it's connected to whether they have a file with a particular hash and, if one of them does, they send back the whole file. Without a short, unique identifier like a cryptographic hash, that wouldn't be possible. This technique is called [content addressing](/concepts/content-addressing/) — because the content itself is used to form an address, rather than information about the computer and disk location it's stored at.
+
+## Content Identifiers are not file hashes
```
johnnymatthews, line 56:
```suggestion
## Content identifiers are not file hashes
```

docs/concepts/hashing.md
```diff
@@ -49,6 +49,56 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+
+Hash functions are widely used as to check for file integrity. A download provider may publish the output of a hash function for a file. This hash is called the checksum of that file. The checksum enables downloaders of the file to verify that the file has not been altered since it was published. This check is done by performing the same hash function that was used to generate the checksum. If the local output matches the checksum, the file was not altered and can be trusted.
```
johnnymatthews, line 58:
```suggestion
Hash functions are widely used as to check for file integrity. A download provider may publish the output of a hash function for a file, often called a _checksum_. The checksum enables users to verify that a file has not been altered since it was published. This check is done by performing the same hash function against the downloaded file that was used to generate the checksum. If that checksum that the user receives from the downloaded file exactly matches the checksum on the website, then the user knows that the file was not altered and can be trusted.
```

docs/concepts/hashing.md
```diff
@@ -49,6 +49,56 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+
+Let us look at a concrete example: When you download an image file for Ubuntu Linux you might see the following `SHA-256` checksum on the Ubuntu website listed for verification purposes:
```
johnnymatthews, line 60:
```suggestion
Let us look at a concrete example. When you download an image file for Ubuntu Linux you might see the following `SHA-256` checksum on the Ubuntu website listed for verification purposes:
```

docs/concepts/hashing.md
```diff
@@ -49,6 +49,56 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+
+```
+0xB45165ED3CD437B9FFAD02A2AAD22A4DDC69162470E2622982889CE5826F6E3D ubuntu-20.04.1-desktop-amd64.iso
+```
+
+You can now verify the file integrity after the download of the file:
```
johnnymatthews, line 66:
```suggestion
After downloading the Ubuntu image, you can verify the integrity of the file by hashing the file to make sure the checksums match:
```

docs/concepts/hashing.md
```diff
@@ -49,6 +49,56 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+
+```
```
johnnymatthews, line 68:
```suggestion
```shell
```

docs/concepts/hashing.md
```diff
@@ -49,6 +49,56 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+echo "b45165ed3cd437b9ffad02a2aad22a4ddc69162470e2622982889ce5826f6e3d *ubuntu-20.04.1-desktop-amd64.iso" | shasum -a 256 --check
+
+ubuntu-20.04.1-desktop-amd64.iso: OK
+```
+
+Let us compare the checksum provided with the download of our file with the hash included in the CID. We can use `ipfs add` with the `-n`switch to create the CID for a file without adding it to IPFS:
```
johnnymatthews, line 74:
```suggestion
If we add the `ubuntu-20.04.1-desktop-amd64.iso` file to IPFS we receive a hash as an output:
```

It's probably best to keep this as simple as possible. Taking out the `-n` option helps keep things simple.

docs/concepts/hashing.md
```diff
@@ -49,6 +49,56 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+
+```
```
johnnymatthews, line 76:
```suggestion
```shell
```

docs/concepts/hashing.md
```diff
@@ -49,6 +49,56 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+ipfs add ubuntu-20.04.1-desktop-amd64.iso -n
```
johnnymatthews, line 77:
```suggestion
ipfs add ubuntu-20.04.1-desktop-amd64.iso
```

docs/concepts/hashing.md
```diff
@@ -49,6 +49,56 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+
+added QmPK1s3pNYLi9ERiq3BDxKa4XosgWwFRQUydHUtz4YgpqB ubuntu-20.04.1-desktop-amd64.iso
+ 2.59 GiB / 2.59 GiB [==========================================================================================] 100.00%
+```
+
+The string `QmPK1s3pNYLi9ERiq3BDxKa4XosgWwFRQUydHUtz4YgpqB` returned by the `ipfs add` command is the content identifier of the file `ubuntu-20.04.1-desktop-amd64.iso`. The content identifier contains various information in on human readable format. We can utilize the [CID Inspector](https://cid.ipfs.io/) in order to see what the CID includes. The actual hash is listed under `DIGEST (HEX)`:
```
johnnymatthews, line 83:
```suggestion
The string `QmPK1s3pNYLi9ERiq3BDxKa4XosgWwFRQUydHUtz4YgpqB` returned by the `ipfs add` command is the content identifier of the file `ubuntu-20.04.1-desktop-amd64.iso`. We can utilize the [CID Inspector](https://cid.ipfs.io/) to see what the CID includes. The actual hash is listed under `DIGEST (HEX)`:
```

docs/concepts/hashing.md
```diff
@@ -49,6 +49,56 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+
+```
+NAME: sha2-256
+BITS: 256
+DIGEST (HEX): 0E7071C59DF3B9454D1D18A15270AA36D54F89606A576DC621757AFD44AD1D2E
+```
+
+We can now check if the hash contained in the files CID actually equals the checksum for the file:
```
johnnymatthews, line 91:
```suggestion
We can now check if the hash contained in the CID equals the checksum for the file:
```

docs/concepts/hashing.md
```diff
@@ -49,6 +49,56 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+
+```
```
johnnymatthews, line 93:
```suggestion
```shell
```

docs/concepts/hashing.md
```diff
@@ -49,6 +49,56 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+echo "0E7071C59DF3B9454D1D18A15270AA36D54F89606A576DC621757AFD44AD1D2E *ubuntu-20.04.1-desktop-amd64.iso" | shasum -a 256 --check
+
+ubuntu-20.04.1-desktop-amd64.iso: FAILED
+shasum: WARNING: 1 computed checksum did NOT match
+```
+
+::: tip
+Please note that names of hash functions are not used consistently.`SHA-2`, `SHA-256` or `SHA-256 bit` are all referring to the same hash function.
```
johnnymatthews, line 101:
```suggestion
The names of hash functions are not used consistently.`SHA-2`, `SHA-256` or `SHA-256 bit` all refer to the same hash function.
```

johnnymatthews (changes requested):

johnnymatthews:
This is a great first draft @alexmmueller! Something that's important to expand upon is that the hash we get when using a CID inspector is the hash of the root of a tree that includes all the data in your file.

> I decided to not add another hint on how input changes change the output of a has function as this is IMO not really related to the task at hand and already captured quite well in the existing documentation. - @alexmmueller 

That's a fair point. However, it would still be useful to have a straightforward mathematical example that shows the reader how a hashing function turns a file into a hash. [It's the 3rd task in the original issue](https://github.com/ipfs/ipfs-docs/issues/430).

I'm looking forward to reviewing this once you're draft is complete! Let me know if you've got any further questions.

alexmmueller:
Thank you for your feedback @johnnymatthews. I will add the requested changes.

alexmmueller:
@johnnymatthews I included more information on how the DAG actually works, but ended up linking to the respective concept article and IPFS Primer tutorial. I thought those resources cover the topic very deeply and did not want to duplicate information.  Does this work for you?

I still have doubts on the value of going through a trivial hash function example and did not yet succeed in finding a good hash function candidate for such an example Do you have an immediate idea? 

An alternative idea would be to just extend the section on hashing function properties and explain there what **uncorrelated** means by adding an example how the change of one digit changes the SHA-1 hash. What do you think?

alexmmueller:
@johnnymatthews Any comments from your end?

docs/concepts/hashing.md
```diff
@@ -49,6 +49,60 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+We can now check if the hash contained in the CID equals the checksum for the file:
+
+```shell
+echo "0E7071C59DF3B9454D1D18A15270AA36D54F89606A576DC621757AFD44AD1D2E *ubuntu-20.04.1-desktop-amd64.iso" | shasum -a 256 --check
+
+ubuntu-20.04.1-desktop-amd64.iso: FAILED
+shasum: WARNING: 1 computed checksum did NOT match
+```
+
+As we can see the hash included in the CID does NOT match the hash of the input file `ubuntu-20.04.1-desktop-amd64.iso`. To understand what the hash included in the CID actually is, we have to understand how IPFS stores files. IPFS uses a so called Distributed Acyclic Graph (DAG) to keep track of all the data stored in IPFS. A CID identifies one specific node in this graph. This identifier is the result of hashing the node's contents — any opaque payload carried by the node and the list of identifiers of its children — using a cryptographic hash function like `SHA256`. To learn more about wow the DAG is used within IPFS to address content is explained in detail in the article covering the [Merkle DAG](/concepts/merkle-dag/).
```
johnnymatthews, line 104:
```suggestion
As we can see, the hash included in the CID does NOT match the hash of the input file `ubuntu-20.04.1-desktop-amd64.iso`. To understand what the hash contained in the CID is, we must understand how IPFS stores files. IPFS uses a [distributed acyclic graph (DAG)](/concepts/merkle-dag/) to keep track of all the data stored in IPFS. A CID identifies one specific node in this graph. This identifier is the result of hashing the node's contents using a cryptographic hash function like `SHA256`.
```

docs/concepts/hashing.md
```diff
@@ -49,6 +49,60 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+
+::: tip
+The [Tutorial: Merkle Trees and the IPFS DAG](https://dweb-primer.ipfs.io/ipfs-dag) inlcudes in-depth lessons coveing the IPFS DAG.
+:::
```
johnnymatthews, line 108:
```suggestion
```

I'm reluctant to include any links to IPFS Primer docs. They're pretty outdated. This particular link was last updated 12 months ago.

johnnymatthews:
Sorry for the delay @alexmmueller. I've added some comments and suggestions. I think you're right, while having a visual example of how hash functions work, it's not strictly necessary.

docs/concepts/hashing.md
```diff
@@ -49,6 +49,60 @@ Cryptographic hashes come with a couple of very important characteristics:
...
+If we add the `ubuntu-20.04.1-desktop-amd64.iso` file to IPFS we receive a hash as an output:
+
+```shell
+ipfs add ubuntu-20.04.1-desktop-amd64.iso
+
+added QmPK1s3pNYLi9ERiq3BDxKa4XosgWwFRQUydHUtz4YgpqB ubuntu-20.04.1-desktop-amd64.iso
+ 2.59 GiB / 2.59 GiB [==========================================================================================] 100.00%
+```
+
+The string `QmPK1s3pNYLi9ERiq3BDxKa4XosgWwFRQUydHUtz4YgpqB` returned by the `ipfs add` command is the content identifier of the file `ubuntu-20.04.1-desktop-amd64.iso`. We can utilize the [CID Inspector](https://cid.ipfs.io/) to see what the CID includes. The actual hash is listed under `DIGEST (HEX)`:
```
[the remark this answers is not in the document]
alexmmueller: I think a screenshot would be good as we use only text in this section up to now and a visual would help to keep the attention level. Will add!

johnnymatthews:
Are you happy with this PR @alexmmueller, or do you still have some changes to make? If you're happy, can you click the **Ready for review** button, and I'll give everything one last read :)

alexmmueller:
Hey @johnnymatthews! I will add the screenshot you suggested before and will file the PR for review soon.

johnnymatthews (approved):
This looks great! Thanks for your work on this @alexmmueller!

PR #658, opened 2021-02-22
docs: make it clear that _dnslink should be used

lidel:
This PR removes vagueness around where TXT record should be located:

- Resolving TXT records on main name works only because it is supported for legacy reasons
- putting TXT record on  `_dnslink` it is both better for opsec and performance (no unrelated TXT results)

### DIFF
--- a/docs/concepts/dnslink.md
+++ b/docs/concepts/dnslink.md
@@ -11,25 +11,26 @@ DNSLink uses [DNS TXT](https://en.wikipedia.org/wiki/TXT_record) records to map 
 A DNSLink address looks like an [IPNS](/guides/concepts/ipns) address, but it uses a domain name in place of a hashed public key:
 
 ```
-/ipns/myexampledomain.org
+/ipns/example.org
 ```
 
 Just like normal IPFS addresses, they can include links to other files — or other types of resources that IPFS supports, like directories and symlinks:
 
 ```
-/ipns/myexampledomain.org/media/
+/ipns/example.org/media/
 ```
 
 ## Publish Using a Subdomain
 
-While you can publish the TXT record to the exact domain if you so wish, it can be more advantageous to publish DNSLink records using a special subdomain called `_dnslink`. This enables you to improve the security of an automated setup or delegate control over your DNSLink records to a third party without giving away full control over the original DNS zone.
+Publish the TXT record using a special subdomain called `_dnslink`. 
 
-For example, [`docs.ipfs.io`](https://docs.ipfs.io) does not have a TXT record, but the page still loads
-because a TXT record exists for `_dnslink.docs.ipfs.io`. If you look up the DNS records for `_dnslink.docs.ipfs.io`, you'll see its DNSLink entry:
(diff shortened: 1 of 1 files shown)


johnnymatthews (approved):

aschmahmann:
This seems reasonable to me although I am by no means a DNS expert. Unless @Stebalien has any thoughts/concerns about this seems to me like a reasonable thing to go ahead with.

Stebalien:
Only @jbenet knows the truth here. I _believe_ allowing this on the root domain _may_ have been to handle cases where some domain registrars charge per subdomain etc.

lidel:
Charging per subdomain could be the case in the past, but is not an issue these days, we have many _free_ alternatives.   

Based on my spelunking and [notes like this FAQ](https://dnslink.io/#why-not-just-use-_dnslinkdomain-always-instead-of-both-_dnslinkdomain-and-domain) I believe the root domain is pure legacy: it was supported first, and then issues with `ALIAS` and `CNAME` were identified + opsec concerns with credentials for root domain being stored in  CI scripts.

I see no harm in merging this.

lidel:
I've removed use of "domain" in DNSLink docs to reduce confusion, making the doc about "hostnames" and "DNS names", which should be  less ambiguous (https://github.com/ipfs/ipfs-docs/pull/658/commits/76b928ae48a5913a9835d361be45b4d291b3ee53).

docs/concepts/dnslink.md
```diff
@@ -1,39 +1,40 @@
...
-because a TXT record exists for `_dnslink.docs.ipfs.io`. If you look up the DNS records for `_dnslink.docs.ipfs.io`, you'll see its DNSLink entry:
+This not only makes DNSLink lookup more efficient by only returning relevant TXT records, but enables you to improve the security of an automated setup or delegate control over your DNSLink records to a third party without giving away full control over the original DNS zone.
 
-```sh
+For example, [`docs.ipfs.io`](https://docs.ipfs.io) loads because a TXT record exists for `_dnslink.docs.ipfs.io`. If you look up the DNS records for `_dnslink.docs.ipfs.io`, you'll see its DNSLink entry:
+
+```console
 $ dig +noall +answer TXT \_dnslink.docs.ipfs.io
-\_dnslink.docs.ipfs.io. 34 IN TXT "dnslink=/ipfs/QmVMxjouRQCA2QykL5Rc77DvjfaX6m8NL6RyHXRTaZ9iya"
+\_dnslink.docs.ipfs.io. 30 IN TXT "dnslink=/ipfs/bafybeieenxnjdjm7vbr5zdwemaun4sw4iy7h4imlvvl433q6gzjg6awdpq"
```
johnnymatthews, line 33:
```suggestion
> \_dnslink.docs.ipfs.io. 30 IN TXT "dnslink=/ipfs/bafybeieenxnjdjm7vbr5zdwemaun4sw4iy7h4imlvvl433q6gzjg6awdpq"
```

docs/concepts/dnslink.md
```diff
@@ -1,39 +1,40 @@
...
 
 ```
 
-## Resolve Using DNSLink
+## Resolve DNSLink Name
```
johnnymatthews, line 37:
```suggestion
## Resolve DNSLink name
```

johnnymatthews (approved):

PR #664, opened 2021-02-24
Fix downloaded filename for linux ipfs-updater

S-o-T:
There is a mistake in a doc for ipfs-updater (linux section): downloaded filename doesn't match the name of file to be unzipped.


johnnymatthews:
Nice catch! I have a feeling that the `darwin` instructions were copied over from the command-line instructions. Thanks for flagging and fixing this one.

PR #679, opened 2021-03-04
Update glossary.md with a table of contents

BigLep:
The goal is make it easier to see the list of terms quickly rather than having to scroll through the page's content.  The toc would be further cleaned up if we remove the letter headings.  I'm good either way on that.

### DIFF
--- a/docs/concepts/glossary.md
+++ b/docs/concepts/glossary.md
@@ -10,7 +10,7 @@ related:
 
 # IPFS glossary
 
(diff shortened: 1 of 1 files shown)


BigLep:
Here is what this looks on localhost.

image

johnnymatthews:
Thanks for the PR! I have a few concerns however.

I'm not a huge fan of adding a table of contents (ToC). Most users find the glossary by using the search bar at the top right of the site, so having a ToC doesn't serve much of a purpose. On-top of that, this ToC is _very_ long, and will only get longer.

image

If the user is simply browsing to find new terms they don't know, then scrolling down the page in it's present form does the job.

I'm closing this PR. If you feel that there's a case to be made for adding a ToC to this page, please leave comments here.

BigLep:
Sounds good and make sense.  I think this comes from me being used to editing mediawiki pages with tables of contents at the top of the page and being trained to look at that to quickly scan a page.  Anyways, sounds good - thanks.

PR #709, opened 2021-03-22
use CIDv1 in NFT guide

### LINKED ISSUE #707: Use CIDv1 in Minty example, not CIDv0.
https://docs.ipfs.io/how-to/mint-nfts-with-ipfs uses CIDv0 in URIs which [does not work everywhere](https://github.com/ipfs-shipyard/ipfs-companion/pull/824) (only Brave supports it).

@yusefnapora mind re-generating example after  https://github.com/yusefnapora/minty/pull/13 is merged, and updating the docs to CIDv1? It will save everyone headache in the long run :pray:

yusefnapora:
This changes the CIDs in the examples for the NFT minting guide to use CIDv1 instead of V0 (`bafy...` instead of `Qm...`), which plays nicer with web browsers and is generally what we want to recommend as "best practice".

@lidel made a [PR to minty](https://github.com/yusefnapora/minty/pull/13) to make the switch there that I just merged in, so this just makes the guide consistent with the code.

closes #707

### DIFF
--- a/docs/how-to/mint-nfts-with-ipfs.md
+++ b/docs/how-to/mint-nfts-with-ipfs.md
@@ -34,7 +34,7 @@ This problem was demonstrated by an artist who [_pulled the rug_](https://cointe
 
 IPFS solves this problem thanks to [Content Addressing][docs-cid]. Adding data to IPFS produces a content identifier (CID) that's directly derived from the data itself and links to the data in the IPFS network. Because a CID can only _ever_ refer to one piece of content, we know that nobody can replace or alter the content without breaking the link.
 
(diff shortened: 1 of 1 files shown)


johnnymatthews (approved):

lidel:
I think we should make sure EVERY URI used in docs  resolve, because if someone pastes it to Brave and gets error, the trust in IPFS is decreased.

Some URIs are broken (outdated filename), for example:
- `ipfs://bafybeihhii26gwp4w7b7w7d57nuuqeexau4pnnhrmckikaukjuei2dl3fq/my-nft.jpeg` returns `ipfs resolve -r /ipfs/bafybeihhii26gwp4w7b7w7d57nuuqeexau4pnnhrmckikaukjuei2dl3fq/my-nft.jpeg: no link named "my-nft.jpeg" under bafybeihhii26gwp4w7b7w7d57nuuqeexau4pnnhrmckikaukjuei2dl3fq`

@yusefnapora are you still in the zone to fix this? :pray:

PR #710, opened 2021-03-24
docs: add recommendation for ipfs.io/ipfs/CID

olizilla:
Start the address-ipfs-on-the-web article with a clear recommendation for using https://ipfs.io/ipfs/<CID> when linking to content on IPFS.

Currently `ipfs add` still returns a cid v0, so it can't trivially be used in a subdomain, and many usecases (e.g NFTs) are not concerned about domain level security isolation as they point to media. They definitly dont want to be using ipfs:// scheme urls, as those only work in tools that are already ipfs aware.

License: MIT
Signed-off-by: Oli Evans <oli@tableflip.io>

### DIFF
--- a/docs/how-to/address-ipfs-on-web.md
+++ b/docs/how-to/address-ipfs-on-web.md
@@ -6,7 +6,17 @@ description: Hands-on guides to using and developing with IPFS to build decentra
 
 # Address IPFS on the Web
 
-This document is a guide to how to address IPFS content paths on the web.
+_How to link to content on IPFS._
+
+```
+https://ipfs.io/ipfs/<CID>
+# e.g
+https://ipfs.io/ipfs/Qme7ss3ARVgxv6rXqVPiikMJ8u2NLgmgszg13pYrDKEoiu
+```
+
+Browsers that support IPFS can redirect these requests to your local IPFS node, while those that don't can fetch the resource from the ipfs.io gateway. 
+
+You can swap out `ipfs.io` for your own http-to-ipfs gateway, but you are then obliged to keep that gateway running _forever_. If your gateway goes down, users with IPFS aware tools will still be able to fetch the content from the IPFS network as long as any node still hosts it, but for those without, the link will be broken. Don't do that.
 
 ## Dweb addressing in brief
 


lidel (changes requested):
Just noting that recommending `ipfs.io` is problematic:


- Main problem: ties arbitrary content to IPFS project website (https://github.com/protocol/bifrost-infra/issues/178)
  We are moving our GUIs to copy linksto `dweb.link` instead. This undoes this a bit.
- Side problem: bit worried recommending path gateway may undo the work we've  put into bringing site isolation to gateways.  In future version of go-ipfs I want to effectively disable JS/cooikies/storage on path gateways (https://github.com/ipfs/in-web-browsers/issues/157), so websites are forced to used subdomains, so this may not be as big of a problem as it is right now.

Would  it be ok to replace it with `dweb.link`?
Path variant will still work (https://docs.ipfs.io/how-to/address-ipfs-on-web/#automatic-leverage-the-gateway-in-go-ipfs) and we don't use the same origin as `ipfs.io` website.

johnnymatthews (approved):

olizilla:
We `could` make this doc recommend `https://dweb.link/ipfs/<cid>` but this would be literally the first I've heard of it, and would need to be accompanied with a significant effort to upgrade all our brains with that knowledge. How about we land this as-is, as it captures what has been best practice for years, and then do a PR push to get everyone on the dweb.link train?

_I am excited that path based redirecting works on dweb.link. That is cool._ 

> Main problem: ties arbitrary content to IPFS project website

I hear you, but that is only a motivation for moving the ipfs.io website to another domain. I'm working on the assumption that ipfs.io should remain a path based gateway forever, or else we become _the breaker of links._

> Side problem: bit worried recommending path gateway may undo the work we've put into bringing site isolation to gateways.

also reasonable, but i would argue that you've done the work of making subdomain gateways viable, but we as an org now have to come up with a co-ordinated campaign to make them something the wider community understands and uses. Anyone who has used IPFS in the base 5 years will likely have some awareness of ipfs.io/ipfs/cid as a thing, but a very small percentage of current users will know about the subdomain flavour... _at least partly because ipfs add does not produce bas32 cids yet, so there is a non-trivial amount of knowledge required to work with them_

olizilla:
@lidel for your radar, re: site islolation with dweb.link – https://html.spec.whatwg.org/multipage/origin.html#relaxing-the-same-origin-restriction
in subdomain land two subdomains can contrive to share an origin by setting `document.domain`. So in theory any subdomain`*.ipfs.dweb.link` could attempt to share an origin with`dweb.link`. It's depreacted, but still shows up as fully supported across the board. https://caniuse.com/?search=document.domain

Some good news though, it can be prevented with CORS headers, and
> Finally, setting document.domain does not change the origin used for origin-checks by some Web APIs, preventing sub-domain access via this mechanism. Affected APIs include (but are not limited to): Window.localStorage, IndexedDB_API, BroadcastChannel, SharedWorker .

lidel (approved):
@olizilla  iirc `document.domain` does not work on dweb.link (because it is on https://publicsuffix.org list), but even when its allowed, it requires mutual opt-in (means, you can't  force some website to share origin unless it agrees to do so).  We may harden it even further as part of https://github.com/ipfs/in-web-browsers/issues/157. 

As for this PR, I no longer feel that this will be a step backwards for subdomain gateways. 
https://github.com/ipfs/in-web-browsers/issues/157 will provide hard incentives to use them for websites, which removes my concerns. 

@johnnymatthews feel free to merge this.

PR #724, opened 2021-04-01
file-systems/unixfs: fix default leaf format.

raulk:


docs/concepts/file-systems.md
```diff
@@ -285,9 +285,9 @@ When an object is added to IPFS, it is chunked up into smaller parts, each part
 
 The leaf format takes two format options, UnixFS leaves and raw leaves:
 
-- The UnixFS leaves format adds a data wrapper on newly added objects to produce UnixFS leaves with additional data sizes. This wrapper is used to determine whether newly added objects are files or directories.
+- The UnixFS leaves format (default) adds a data wrapper on newly added objects to produce UnixFS leaves with additional data sizes. This wrapper is used to determine whether newly added objects are files or directories.
```
lidel, line 288:
```suggestion
- The UnixFS leaves format (default for CIDv0) adds a data wrapper on newly added objects to produce UnixFS leaves with additional data sizes. This wrapper is used to determine whether newly added objects are files or directories.
```

docs/concepts/file-systems.md
```diff
@@ -285,9 +285,9 @@ When an object is added to IPFS, it is chunked up into smaller parts, each part
...
 
-- The raw leaves format is the default format on IPFS where nodes output from chunking will be raw data from the file with a CID type of 'raw'. This is mainly configured for backward compatibility with formats that used a UnixFS Data object.
+- The raw leaves format (experimental, likely future default) on IPFS where nodes output from chunking will be raw data from the file with a CID type of 'raw'. This is mainly configured for backward compatibility with formats that used a UnixFS Data object.
```
lidel, line 290:
```suggestion
- The raw leaves format (default for CIDv1 created with `ipfs add --cid-version 1`, soon to become the global default) on IPFS where nodes output from chunking will be raw data from the file with a CID codec of 'raw'. This is mainly configured for backward compatibility with formats that used a UnixFS Data object.
```

lidel (commented):
I think it's useful to note that raw leaves are enabled by default when user opts-in to CIDv1 during the import, and note that will become the default soon. 

Feel free to rephrase below suggestions.

raulk:
Thanks @lidel, done!

achingbrain:
> note that will become the default soon

Is there a timeline for this?  Last time this came up I think @Stebalien had some ideas around introducing 'import profiles' that would apply raw leaves, cidv1, etc, rather than changing the defaults.

Stebalien (approved):

Stebalien:
> Last time this came up I think @Stebalien had some ideas around introducing 'import profiles' that would apply raw leaves, cidv1, etc, rather than changing the defaults.

Honestly, at this point, I'd just change the default and document it in the release notes. If it's _just_ one flag, that shouldn't be a big issue. As we make more and more changes, we can go back and add a profiles flag.

I'm also happy to do that now, but I don't want to block on that.

docs/concepts/file-systems.md
```diff
@@ -285,9 +285,9 @@ When an object is added to IPFS, it is chunked up into smaller parts, each part
 
 The leaf format takes two format options, UnixFS leaves and raw leaves:
 
-- The UnixFS leaves format adds a data wrapper on newly added objects to produce UnixFS leaves with additional data sizes. This wrapper is used to determine whether newly added objects are files or directories.
+- The UnixFS leaves format (default for CIDv0) adds a data wrapper on newly added objects to produce UnixFS leaves with additional data sizes. This wrapper is used to determine whether newly added objects are files or directories.
```
johnnymatthews, line 288:
```suggestion
- The UnixFS leaves format adds a data wrapper on newly added objects to produce UnixFS leaves with additional data sizes. This wrapper is used to determine whether newly added objects are files or directories. This format is the default for CIDv0.
```

docs/concepts/file-systems.md
```diff
@@ -285,9 +285,9 @@ When an object is added to IPFS, it is chunked up into smaller parts, each part
...
 
-- The raw leaves format is the default format on IPFS where nodes output from chunking will be raw data from the file with a CID type of 'raw'. This is mainly configured for backward compatibility with formats that used a UnixFS Data object.
+- The raw leaves format (default for CIDv1 created with `ipfs add --cid-version 1`, soon to become the global default) on IPFS where nodes output from chunking will be raw data from the file with a CID codec of 'raw'. This is mainly configured for backward compatibility with formats that used a UnixFS Data object.
```
johnnymatthews, line 290:
```suggestion
- The raw leaves format on IPFS where nodes output from chunking will be raw data from the file with a CID codec of 'raw'. This is mainly configured for backward compatibility with formats that used a UnixFS Data object. This format is the default for CIDv1 created with `ipfs add --cid-version 1`, soon to become the global default.
```

johnnymatthews (approved):

PR #729, opened 2021-04-02
NFT best practices

yusefnapora:
This is an early draft of the NFT "best practices" documentation that we're working on. 

It's not ready for review yet (I'll be adding things over the weekend and Monday morning), but I'm putting it up as draft now for visibility, as I'm running out of steam for the day.

### DIFF
--- a/docs/how-to/best-practices-for-nft-data.md
+++ b/docs/how-to/best-practices-for-nft-data.md
@@ -38,7 +38,7 @@ Here's an example of a full IPFS URI: `ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf
 
 IPFS URIs are the canonical representation for an IPFS link, since the `ipfs` scheme makes it clear and unambiguous that the CID refers to content on IPFS and not some other system. To produce an IPFS URI, simply prefix a CID string with the static string `ipfs://`.
 
-You can also include filenames inside the path component of an IPFS URI. For example, if you've stored your token's metadata on IPFS wrapped in a directory, your URI might be: `ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi/metadata.json`
+You can also include filenames inside the path component of an IPFS URI. For example, if you've stored your token's metadata on IPFS wrapped in a directory, your URI might be: `ipfs://bafybeibnsoufr2renqzsh347nrx54wcubt5lgkeivez63xvivplfwhtpym/metadata.json`
 
 We recommend using an IPFS URI to link from your smart contract to any external data stored using IPFS, including any _metadata_ that describes and contextualizes the token.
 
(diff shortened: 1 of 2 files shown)


mikeal:
I’d cut the v0, v1 and base encoding information. The “right thing” happens when you use `cid.toString()` so we can just skip right to that part. As long as we link to the CID docs they can dig into these details if they wish, but this isn’t information they need in order to get the NFT minted.

yusefnapora:
I realized last night that recommending binary CIDs on-chain is problematic - the `tokenURI` function needs to return a string, so if you're storing binary CIDs or raw hashes, you'd have to do the base32 conversion on-chain, which would probably cancel out the cost savings.

yusefnapora:
I think this is ready for review. Here's the [latest deploy preview](https://bafybeihxd4opm447ukg53ynlmg456q6vstbr6qwak4unxkmr26ze6ac7gi.on.fleek.co/how-to/best-practices-for-nft-data/#metadata)

docs/how-to/best-practices-for-nft-data.md
```diff
@@ -0,0 +1,160 @@
...
+
+This guide is aimed at developers building NFT platforms and other tools, and it's focused on how to format your data and link to it for the best long-term results. If you're looking for details about smart contract interactions and how token minting works, head over to our [guide to minting NFTs with IPFS][docs-mint-nfts], where we go over the whole process from end-to-end using an Ethereum test network.
+
+## Types of IPFS links and when to use them
+
+There are a few different ways to refer to data on IPFS, each of which is best suited to different use cases.
+
+### CID
+
+First, there's the "raw" CID, which uniquely identifies a specific piece of content. CIDs can be stored and sent over the network in a compact binary form, but they're represented as strings of random-seeming characters when displayed to users. Here's an example:
```
mikeal, line 21: We can cut the first sentence about “raw.” The second sentence is a great initial statement and we don’t need to get into codecs.

docs/how-to/best-practices-for-nft-data.md
```diff
@@ -0,0 +1,160 @@
...
+
+```
+bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
+```
+
+::: tip
+You can learn more about CIDs in our [guide to Content Addressing][docs-cid], or by following the [interactive tutorials on ProtoSchool][protoschool-cid].
+:::
+
+Because it's the most compact form of IPFS link, the CID is a good fit for storing directly on the blockchain. We recommend storing CIDs as strings, in the base 32 encoding as shown in the example above.
```
mikeal, line 31: We don’t need to get into base encoding, our implementations return the preferred string form by default and the linked docs go into greater detail.

yusefnapora:
Per @momack2's feedback on the related blog post, I'm going to remove the advice to store CIDs on chain without the URI scheme. It's too small of an optimization to justify the confusing distinction in a doc like this, now that we're not recommending binary CIDs.

jessicaschilling:
Per OOB thread from @momack2 and @atopal, and at the request of @emilymvaughan, merging.

PR #739, opened 2021-04-13
add link to instructions step

mburns:
This wording is similar to the Windows instructions, and provides a clickable link to the page we want them to go to.

I like clickable links.


johnnymatthews (approved):
Nice :)

PR #741, opened 2021-04-13
fix macOS download reference and make first step consistent

tommerkle1:
Made the download step consistent across all platforms. I noticed this first when looking at the macOS directions where it was pointing me to a nonexistent github repo to find the .dmg


johnnymatthews:
Yeah that cleans things up a bit! Thanks for your help! Merging into staging.

PR #745, opened 2021-04-18
Fix broken link in how-to/pin-files/

brianstrauch:
https://docs.ipfs.io/how-to/pin-files/#using-a-pinning-service does not exist.
Redirect to https://docs.ipfs.io/how-to/work-with-pinning-services/


johnnymatthews:
Nice find, thanks! Looks like this got missed in a recent pinning service update. Merging into staging.

PR #747, opened 2021-04-19
fix italic md formatting

natesales:


johnnymatthews:
Nice find, thanks! Merging into staging.

PR #748, opened 2021-04-19
Giving an example of how to set the X-Ipfs-Path

CSDUMMI:
There was some confusion as to what format the header should be, so to
avoid these in the future, I added an example to this page.


lidel:
Thank you @CSDUMMI !

I think we need to have separate examples for `/ipfs` and `/ipns`.
Let's continue in https://github.com/ipfs/ipfs-docs/pull/749

PR #761, opened 2021-04-23
Document that Brave also supports x-ipfs-path

da2x:


docs/how-to/companion-x-ipfs-path-header.md
```diff
@@ -7,6 +7,8 @@ description: Learn more about how to use "x-ipfs-path" headers in IPFS Companion
 
 IPFS Companion can redirect traditional HTTP requests to IPFS if the `x-ipfs-path` response header is provided.
 
+Additionally, the Brave browser will display an “Open using IPFS” button on the address bar when this header is returned by a tab’s main document frame.
```
lidel, line 10:
```suggestion
Additionally, some browser vendors like [Brave](https://brave.com/ipfs-support/) may display an “Open using IPFS” button on the address bar when this header is returned for the root document in the current tab.
```

lidel (approved):
lgtm, but would make it more open-ended (Brave is the first, but there will be more) + point at more details about Brave for better SEO/discovery

PR #769, opened 2021-04-29
recommend CIDv1 in NFT best practices

yusefnapora:
This adds a recommendation to use CIDv1 for NFTs, along with a quick explanation of how to tell the difference and how to get v1 cids from `ipfs add`.

cc @autonome

### DIFF
--- a/docs/how-to/best-practices-for-nft-data.md
+++ b/docs/how-to/best-practices-for-nft-data.md
@@ -24,11 +24,16 @@ CIDs uniquely identify a piece of content. A CID can be stored and sent over the
 bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
 ```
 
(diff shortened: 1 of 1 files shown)


docs/how-to/best-practices-for-nft-data.md
```diff
@@ -24,11 +24,16 @@ CIDs uniquely identify a piece of content. A CID can be stored and sent over the
 bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
 ```
 
+There are two versions of CIDs used by IPFS. The example above is a version 1 CID (or CIDv1), and it has some advantages over the older "version 0" format, especially when viewing IPFS content on the web using an IPFS gateway. It's best to use version 1 CIDs for addressing NFT data. In javascript, version 1 CIDs are already the default, so if you're using js-ipfs, you're all set. If you're using the go-ipfs command line to add your data to IPFS, you can use the `--cid-version=1` flag when running `ipfs add`.
```
lidel, line 27: > In javascript, version 1 CIDs are already the default, so if you're using js-ipfs, you're all set.

I don't believe this is true:

```
echo "can i has cidv1?" |  jsipfs add --only-hash                                                                                                                                            
added QmQSv22dCdwrtrU84nTCUhJYyeyFfQeyxMdDTcyGzQzreQ QmQSv22dCdwrtrU84nTCUhJYyeyFfQeyxMdDTcyGzQzreQ
```

lidel (changes requested):

docs/how-to/best-practices-for-nft-data.md
yusefnapora: Thanks, I don't know why I thought so...

lidel (approved):

PR #775, opened 2021-05-06
fix example CIDs in NFT guide

### LINKED ISSUE #770: ipfs:// URIs on NFT tutorial do not resolve
A friend  noted that copying and pasting "permanent" URIs from https://docs.ipfs.io/how-to/mint-nfts-with-ipfs/ to Brave returns error for most of them:

- `ipfs://bafybeihhii26gwp4w7b7w7d57nuuqeexau4pnnhrmckikaukjuei2dl3fq/my-nft.jpeg`
- `ipfs://bafybeihhii26gwp4w7b7w7d57nuuqeexau4pnnhrmckikaukjuei2dl3fq/nft-image.png`

I believe it is extremely important that every CID and  `ipfs://` URI we use in docs resolves.

@yusefnapora I guess swapping filenames to `ticket.txt` would do the trick, but then we have "image" named ".txt" which is awkward :)

Do we have any real NFT created since then that we could reuse in examples in that article?
If not, we could use `ipfs://bafybeidfjqmasnpu6z7gvn7l6wthdcyzxh5uystkky3xvutddbapchbopi/no-time-to-explain.jpeg` (already used on https://docs.ipfs.io/how-to/best-practices-for-nft-data/#types-of-ipfs-links-and-when-to-use-them)

yusefnapora:
This fixes the example URIs in the NFT Guide to close #770.

I just updated the broken examples to `ipfs://bafybeidlkqhddsjrdue7y3dy27pu5d7ydyemcls4z24szlyik3we7vqvam/nft-image.png`, which is a pinned copy of the IPFS logo.

Thanks for reporting that @lidel, and sorry about that! I remember you made the point about invalid URIs before we published, but I must missed these.

### DIFF
--- a/docs/how-to/mint-nfts-with-ipfs.md
+++ b/docs/how-to/mint-nfts-with-ipfs.md
@@ -34,7 +34,7 @@ This problem was demonstrated by an artist who [_pulled the rug_](https://cointe
 
 IPFS solves this problem thanks to [Content Addressing][docs-cid]. Adding data to IPFS produces a content identifier (CID) that's directly derived from the data itself and links to the data in the IPFS network. Because a CID can only _ever_ refer to one piece of content, we know that nobody can replace or alter the content without breaking the link.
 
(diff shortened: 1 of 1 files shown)


lidel (approved):
Thanks, and no worries, it is easy to miss this without any automation :)

@johnnymatthews @olizilla do you know if we have a CI setup that finds CIDs / URIs in text files and checks if they resolve? 
If not.. I feel such github action would be super useful across all our docs sites and blogs.
Should I fill an issue? (Where should this live? Ecosystem?)

johnnymatthews:
> do you know if we have a CI setup that finds CIDs / URIs in text files and checks if they resolve?

We do not, as far as I know. You can raise an issue either here, or in IPFS Docs. Then, whatever we use to fix the issue, we can port over to all the other VuePress-based websites.

PR #780, opened 2021-05-18
Update hombrew command

unbeatable-101:
Using `brew cask install ipfs` doesn't work anymore, you need to do `brew install ipfs --cask`


johnnymatthews:
Thanks for the update!

PR #790, opened 2021-06-04
In api.md, mention disabling of window.ipfs.

toolness:
The `window.ipfs` object was disabled in late 2019, but the documentation still mentions it as an available option.  Rather than removing mention of it entirely, this change mentions that it used to exist, and links to the rationale for why it was disabled.  I decided to do this because I am guessing that some people who might be visiting this page may have once used or heard about `window.ipfs` and may be surprised if it were not mentioned at all.


johnnymatthews:
That's a great catch! Thanks for filing this PR!

PR #803, opened 2021-06-16
fixing the wrong file name

sashrika:
We are asked to create the txt file with the name flight-to-the-moon. There is no file called ticket.txt. 
I renamed and fixed it.


johnnymatthews:
Ahh, nice catch. Thanks!

PR #814, opened 2021-06-23
Add reference to direct js-ipfs video support

bneijt:
Add information on how you can directly stream video using js-ipfs and javascript MediaSource API.


BigLep:
@lidel : (low priority) this feels like more of a blog post than something we want to keep committed in our docs website.  That said, I don't have a full working model of the scope of our docs.  Do you have have an opinion?

bneijt:
Sound reasonable to me, so I'll close this pr. Thank you for taking the time to check it out.

PR #826, opened 2021-07-03
[how-to] Using ENS+IPFS to get an ENS domain (ciskip)

Undigon:
Today I found out ENS uses ENS, and the Metamask+IPFS Companion duo seems to work fine so I added it while preserving the traditional method as an option.

Also deleted a seemingly left-over word and a typo from README.md

Best regards.


johnnymatthews (approved):

johnnymatthews:
Thanks for the help!

PR #835, opened 2021-07-12
Update content-addressing.md

cawfree:
Corrected JavaScript snippet to provide a `base32` string parameter to the call to `toString()` when creating a `v1` CID string using [`cids`](https://github.com/multiformats/js-cid).


johnnymatthews:
Thanks!

PR #838, opened 2021-07-13
host git style repo: add removing of packs + avoid copying packs

RubenKelevra:


johnnymatthews:
Thanks for this PR @RubenKelevra. Before I merge this in, what's the perceived benefit of moving the packs rather than copying them over? Just an attempt to avoid duplication?

RubenKelevra:
> Thanks for this PR @RubenKelevra. Before I merge this in, what's the perceived benefit of moving the packs rather than copying them over? Just an attempt to avoid duplication?

Well you can delete the packs after extracting them.

Moving data is nearly instant and atomic on most filesystems.

I had a large repo with around 3 GB of packs which took a considerable amount of time to copy, just to delete the data afterwards on both locations.

johnnymatthews:
> I had a large repo with around 3 GB of packs which took a considerable amount of time to copy, just to delete the data afterwards on both locations.

That's fair. Thanks for the clarification!

Merging.

PR #839, opened 2021-07-16
Wrong case resulting in not scrolling.

namelessperson0:
"Ubuntu" with capital "U" does not scroll to the section. Hence changed


johnnymatthews:
Nice find! Thanks for the fix!

PR #845, opened 2021-07-25
Escaped apostrophe in facts list

jeremyfritzen:
Apostrophes in facts list raises an error, preventing the script from running correctly (facts not displayed). Escaping thse characteres fixes the issue.


johnnymatthews:
Nice! Thanks for catching that one.

PR #858, opened 2021-08-10
document boolean values

asah:
Didn't see this documented anywhere...

### DIFF
--- a/docs/reference/http/api.md
+++ b/docs/reference/http/api.md
@@ -55,7 +55,7 @@ Arguments are added through the special query string key "arg":
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/reference/http/api.md
```diff
@@ -55,7 +55,7 @@ Arguments are added through the special query string key "arg":
 }
 ```
 
-Note that it can be used multiple times to signify multiple arguments.
+Note that it can be used multiple times to signify multiple arguments. Boolean values (type `bool`) may have the value `true` or `false`.
```
johnnymatthews, line 58:
```suggestion
Note that it can be used multiple times to signify multiple arguments. Boolean `bool` values may have the value `true` or `false`.
```

johnnymatthews:
That's a fair point to make. Thanks for the addition!

PR #861, opened 2021-08-11
feat: fixed color contrast and added ipfs styles

### LINKED ISSUE #293: Insufficient color contrast for text links
In working on a fork of this site I discovered that the `$accentColor` that creates the text links on the beta site (both in-line and on the sidebar) doesn't pass contrast checks for accessibility for normal size font on any of the three backgrounds I see it on. Just dropping these reports here (from https://webaim.org/resources/contrastchecker/) in case they're helpful for future accessibility reviews: 

image

image

image

rishabhkeshan:
I have changed the accent color to the new teal used in ipfs-css, along with that imported the fonts and ipfs.css file to the repository.
The new color also passes the contrast check as below.
image

fixes #293


hsanjuan:
Are these fonts used?

rishabhkeshan:
Hey nope, the fonts are currently not required to be used anywhere. Do I remove the fonts? @hsanjuan

hsanjuan:
@rishabhkeshan I'd say yes, remove, unless there is a reason to have them.

rishabhkeshan:
Hey @hsanjuan I have removed the fonts.

package-lock.json
```diff
@@ -1,17438 +1,8 @@
 {
   "name": "ipfs-docs",
   "version": "2.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
hsanjuan, line 4: Why is package.json modified so heavily?

hsanjuan (changes requested):

package-lock.json
rishabhkeshan: I think that happened because I did a ```npm i``` instead of ```npm ci``` so the package-lock got overwritten

rishabhkeshan: @hsanjuan fixed it, sorry for the trouble.

hsanjuan (approved):

PR #863, opened 2021-08-20
Uses `abc.ipfs.eth.link` instead of `abc.ipfs.io`

johnnymatthews:
`ipfs.io` is having issues, so we're moving to `.eth.link` and decentralizing the world.

- Changes `blog.ipfs.io` to `blog.ipfs.eth.link`.
- Changes `docs.ipfs.io` to relative links.
- Removes `legacyURL` field from a bunch of pages. That was for the _super old_ Hugo site we had years ago.
- Changes `dist.ipfs.io` to `dist.ipfs.eth.link`.


johnnymatthews:
The root issue that causes this change has been fixed. Closing.

PR #881, opened 2021-09-15
preserve example consistency

FullByte:
Changed command example path to preserve example consistency (multi-page-first-step)


johnnymatthews:
Good call. Thanks for this PR!

PR #896, opened 2021-09-29
Update pin-files.md

TMoMoreau:
Updated to add the outputs for the commands.

### DIFF
--- a/docs/how-to/pin-files.md
+++ b/docs/how-to/pin-files.md
@@ -1,6 +1,5 @@
 ---
 title: Pin files
-legacyUrl: https://docs.ipfs.io/guides/examples/pinning/
 description: Learn how to pin files in IPFS in order to keep your files and other objects local.
 ---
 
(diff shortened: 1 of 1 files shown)


docs/how-to/pin-files.md
```diff
@@ -12,13 +11,46 @@ While this is nice, sometimes you want to be able to control what you keep aroun
...
 Let's look at this example to explore pinning to your local IPFS node in a bit more depth:
 
-```
-echo "ipfs rocks" > foo
-ipfs add foo
-ipfs pin ls --type=all
-ipfs pin rm <foo hash>
-ipfs pin rm <foo hash>
-ipfs pin ls --type=all
+1. First, create a file called `foo` with the contents `ipfs rocks`:
```
johnnymatthews, line 14: Each tasks comes in a numbered item in a list. Each item starts with `1.`, even if it isn't the first item. Markdown automagically figures out how many items are in the list when it gets converted into HTML, so we don't have to bother writing in the numbers manually.

If a numbered line is followed by a code block (\`\`\`) then this line should end in a colon `:`.

File name and variables, like `foo` and `ipfs rocks`, are surrounded by back-ticks (\`).

docs/how-to/pin-files.md
```diff
@@ -12,13 +11,46 @@ While this is nice, sometimes you want to be able to control what you keep aroun
...
+ipfs pin rm <foo hash>     
+
+> Error: not pinned or pinned indirectly
+
+
+ipfs pin ls --type=all    
+
+> QmQPeNsJPyVWPFDVHb77w8G42Fvo15z4bG2X8D2GhfbSXc recursive
+> QmQy6xmJhrcC5QLboAcGFcAE1tC8CrwDVkrHdEYJkLscrQ indirect
+> ...
```
johnnymatthews, line 53: @TMoMoreau following the structure I've laid out above, can you try and explain what each of these commands does?

johnnymatthews (approved):

PR #904, opened 2021-10-07
chore: go-ipfs 0.10.0

lidel:
I wish we could automate this, or ideally, remove the need for hardcoding version  number in the URLs (having `/latest` dir)


BigLep:
Followup item to automate this: https://github.com/ipfs/ipfs-docs/issues/905

PR #906, opened 2021-10-08
new content: add Handshake 

johnnywu-namebase:
This PR is to include Handshake as another option IPFS users can use for their decentralized domain.


johnnywu-namebase:
@johnnymatthews thanks so much for all the comments and help, apologize if this was a bit messy (first time ever doing a real PR). Lemme know if there's anything else you'd like to see fixed!

johnnywu-namebase:
Thank you for the kind words! Luckily the Handshake instructions were near-identicail to the existing DNS instructions so it was pretty easy to keep the style consistent.

Is it okay to touch the ENS instructions since we're technically a competitor? I don't feel too comfortable going right ahead and doing that, but I would just cut out steps 1 through 10 that are just about how to get an ENS domain and linking out to https://ens.domains/ instead.

Issue #911, opened 2021-10-14
Replace wget with curl in mac CLI install instructions


SuaYoo:
Mac instructions should use `curl` instead of `wget` on the install page: https://docs.ipfs.io/install/command-line/#official-distributions. The updater installation docs correctly uses curl: https://docs.ipfs.io/install/ipfs-updater/#install-updater although I would suggest using `curl -O` instead of repeating the filename with `--output`.

johnnymatthews:
This is a fair point. I agree that we should be consistent across all pages in terms of which tools we use to download content. But what's the reason for macOS using cURL specifically? Why not change the Updater page to use wget?

I'm not trying to start an argument. I'm curious as to why we should encourage macOS users to use cURL, but not Linux users.

SuaYoo:
Macs come with curl installed, but not wget :)

PR #912, opened 2021-10-14
Update macOS CLI download script

SuaYoo:
fixes issue #911


johnnymatthews:
Thanks for the PR, it's really appreciated! I've dropped a question in the [original issue](https://github.com/ipfs/ipfs-docs/issues/911). Let's answer that before we merge this PR.

PR #932, opened 2021-11-06
remove erroneous exclamation point

jrfrantz:
in the example, the person makes a text file with "Hello IPFS".
so `ipfs cat` would return "Hello IPFS". currently there's an exclamation point that came from nowhere that I'm removing.


johnnymatthews:
Ah, nice find! But it looks like the line that is wrong is the 2nd item, where we echo `Hello IPFS!` into `hello.txt`, since all the other items below that feature the `!`. I must have miss-copied that initial echo.

johnnymatthews:
Oh, nevermind. I see what's going on. Including the `!` causes bash to think that we want to ignore the last `"`. I'll take the `!` out of all the examples now.

johnnymatthews:
There we go. Think I got them all.

PR #969, opened 2021-12-09
fix: proper noop when no http-api-updates

lidel:
@guseggert  the error was cosmetic – last step was running when it should not.
Fix from this PR skips it if there is no branch with updates.


filecorgi:

PR #1026, opened 2022-02-12
Remove duplicated info (ciskip)

sauerburger:
The page about [Immutablility](https://docs.ipfs.io/concepts/immutability/) contains almost the same paragraph twice. I think this was added by mistake and I would argue it doesn't really added anything for the reader. However, you might want to keep it anyway.

With this PR, I'm suggesting to remove the duplicated paragraph keeping the more recent version (added in https://github.com/ipfs/ipfs-docs/commit/95fbcea4200353d256b660867ee001bb3c2d0b26). 

(It's kind of ironic that the change to the paragraph itself is of the style "hello world" -> "hello IPFS!")

### DIFF
--- a/docs/concepts/immutability.md
+++ b/docs/concepts/immutability.md
@@ -50,9 +50,6 @@ If we change the content of `B` to `IPFS!`, all the upstream paths will change a
 "hello"    "world"     "hello"    "IPFS!"
 ```
 
-Again, node `B` does not change. It will always refer to the same content, `world`. Node `A` also appears in the new DAG. This is not because we are _keeping_ it; that would imply the location-addressed paradigm. In the content-addressed system, any time someone writes a block with `"hello"`, it will _always_ have CID `A`.
-This is different from location-addressed systems where we could reuse the original buffer and edit the small substring that represents the difference.
-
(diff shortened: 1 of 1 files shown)


Annamarie2019:
From looking at the [9f5bcea](https://github.com/ipfs/ipfs-docs/commit/95fbcea4200353d256b660867ee001bb3c2d0b26), I'm not sure which one of these is the latest version. It shows the duplication as the latest version:

"Again, node B does not change. It will always refer to the same content, world. Node A also appears in the new DAG. This is not because we are keeping it; that would imply the location-addressed paradigm. In the content-addressed system, any time someone writes a block with "hello", it will always have CID A.
This is different from location-addressed systems where we could reuse the original buffer and edit the small substring that represents the difference.

Again, node B does not change. It will always refer to the same content, world. Node A also appears in the new DAG. This does not necessarily mean we copied the memory/buffer that contained the hello string into our new message; that would imply the location-addressed paradigm that focuses on the where and not the what. In a content-addressed system, any time someone writes the string hello it will always have CID A, regardless of whether we copied the string from a previous location or we wrote it from scratch."

The "regardless" phrase seems to be the more eloquent statement. So should we change it like this (yes, it's ironic):
"Again, node B does not change. It will always refer to the same content, world. Node A also appears in the new DAG. This does not necessarily mean we copied the memory/buffer that contained the hello string into our new message; that would imply the location-addressed paradigm that focuses on the where and not the what. In a content-addressed system, any time someone writes the string hello it will always have CID A, regardless of whether we copied the string from a previous location or we wrote it from scratch."

johnnymatthews (approved):
Seams reasonable. Thanks!

PR #1053, opened 2022-02-28
Removes "Example" and redundant last paragraph

Annamarie2019:
"Example" in the title makes it sound like it's an example of the section above it. It's not; it's stand-alone, so I removed the word "example."
In another PR, I had moved the bottom paragraph to the top of the article, so they knew why they were reading it, but I see that the original, now redundant, paragraph is left behind, so I removed it.


filecorgi:
### `docs/concepts/hashing.md`:
- ‘exact same’ might be wordy. Consider a shorter alternative.
  ```
  ...e represented by SHA-1 as: However, the exact same input generates the follo...
                                             ^^^^^^^^^^
  ```
- ‘exactly the same’ might be wordy. Consider a shorter alternative.
  ```
  ...- the same input message always returns exactly the same output hash uncorre...
                                             ^^^^^^^^^^^^^^^^
  ```
- Consider using a different adverb to strengthen your wording.
  ```
  ...l change in the message should generate a completely different hash unique -...
                                             ^^^^^^^^^^^^
  ```

johnnymatthews (approved):

PR #1065, opened 2022-03-08
Pubsub multibase encoding

TMoMoreau:
Documenting and clarifying the requirement of multibase encoding when publishing data to a pubsub topic.

### DIFF
--- a/docs/reference/http/api.md
+++ b/docs/reference/http/api.md
@@ -3803,7 +3803,13 @@ This endpoint returns a `text/plain` response body.
 
 ### cURL Example
 
-`curl -X POST -F file=@myfile "http://127.0.0.1:5001/api/v0/pubsub/pub?arg=<topic>"`
+`curl -X POST -F file=@myfile "http://127.0.0.1:5001/api/v0/pubsub/pub?arg=u<topic>"`
+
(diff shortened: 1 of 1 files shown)


TheDiscordian:
LGTM, I personally really like that you linked to a tool to assist devs in converting their topic string to Base64URL (really handy especially if someone is coming to the docs after wondering why their `pub` isn't working). It is linked to an external tool, not 100% sure how kosher that is, but I'm sure @johnnymatthews could bless us with his opinion/wisdom 🙏.

Tagging @Annamarie2019 hoping this more complete PR will help alleviate some confusion ❤️.

lidel (changes requested):
This is great, but this markdown file is generated, so these changes will get overridden when a new release of go-ipfs ships :( 

You need to update https://github.com/ipfs/ipfs-docs/tree/main/tools/http-api-docs generator and special-case this specific endpoint.

FYSA I filled https://github.com/ipfs/go-ipfs-cmds/issues/224  to add a programatic way to indicate which string fields are expected to be multibase – if someone is able to implement that, then generating docs should be way easier, but it is also fine to special-case in `tools/http-api-docs` for now, if you find it easier.

TheDiscordian:
@lidel thank you so much putting us on the correct path here (even with options)!

TheDiscordian:
Closed PR as per Lidel's message.

Issue #1007, opened 2022-01-26
Document PubSub changes from go-ipfs v0.11


TheDiscordian:
This describes the issue very well: https://github.com/ipfs/go-ipfs/issues/8612

Caused by: https://github.com/ipfs/go-ipfs/issues/8343

Relevant areas: https://docs.ipfs.io/reference/http/api/#api-v0-pubsub-pub && https://docs.ipfs.io/reference/http/api/#api-v0-pubsub-sub (Edit: This actually doesn't effect `sub`, details below!)

Basically, we used to accept regular strings for pubsub topics over the HTTP API, now we require Base64URL encoding **prefixed with a 'u'** (Edit: As outlined below, this isn't *entirely* true, Base64URL is the default/recommended encoding, but technically any multibase should work https://github.com/multiformats/multibase, which is where the `u` comes from). Noting this down in an issue so I don't forget, I spun my wheels on this for a while not realizing this change happened, this is probably a good first issue though.

Also not sure how other APIs have been effected. I believe the CLI handles this for you for example.

Annamarie2019:
We made a change and it's causing bugs. We used to accept regular strings for pubsub topics over the HTTP API, now we require Base64URL encoding prefixed with a 'u'.
These docs need to be fixed: https://docs.ipfs.io/reference/http/api/#api-v0-pubsub-pub && https://docs.ipfs.io/reference/http/api/#api-v0-pubsub-sub
Needs analysis as to whether other APIs are affected.

TMoMoreau:
Fix for the areas specified by @TheDiscordian at PR #1046

Annamarie2019:
@TMoMoreau or @TheDiscordian  Sorry, I'm new to Base64URL encoding and can't find any discussion of it in respect to a prefix. (46 seems to mean `u`, but that doesn't help me any). I don't know how to prefix the `u`. Is there a . or an _ ? Is this doc correct (I bolded what I changed):

### Arguments

- `arg` [string]: Topic to publish to. Required: **yes**. **Requires Base64URL encoding, prefixed with `u`.**

### Request Body

Argument `data` is of file type. This endpoint expects one or several files (depending on the command) in the body of the request as **`umultipart/form-data`**.

As for affecting other APIs, publish is the only one that writes and has a Request Body, so it seemed to me that that's the only one to change. Is that correct thinking? (To read existing stuff, you'd have to call it by whatever it's named.)

TheDiscordian:
[...]

PR #1067, opened 2022-03-09
Adds req for base encoding & points out cURL example

Annamarie2019:
For issue #1007 : 

To the argument: Adds the requirement for multibase encoding

To the cURL example: Adds a `u` before the argument and adds a comment that it's satisfying the argument requirement. @TMoMoreau brought this to my attention in PR #1065 , but I think we should summarize his beautifully word-smithed prose, bc we state it in full in the argument description.

### DIFF
--- a/docs/reference/http/api.md
+++ b/docs/reference/http/api.md
@@ -3784,13 +3784,11 @@ Publish data to a given pubsub topic.
 
 ### Arguments
 
-- `arg` [string]: Topic to publish to. Required: **yes**.
-
-
+- `arg` [string]: Topic to publish to. Required: **yes**. Requires multibase encoding. The default and recommended encoding is `u` for base64url, but you can use any base encoding character. See the Multibase Table at [multibase](https://github.com/multiformats/multibase).
 
 ### Request Body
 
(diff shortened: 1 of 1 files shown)


filecorgi:

johnnymatthews:
This is a sensible suggestion! However, we don't actually make changes to this API documentation manually. This section of the docs `docs.ipfs.io/reference/http/api` is created automatically by the `http-api-docs` tool we have in this repo: https://github.com/ipfs/ipfs-docs/tree/main/tools/http-api-docs

The tool spins up an IPFS node, and then asks the node what it can do. The response from the node is output as markdown. So to update the content, we have to create a PR against https://github.com/ipfs/go-ipfs. Buuut, I'm not quite sure where to make the PR. We'll have to get @aschmahmann or @hsanjuan to point us in the right direction.

Annamarie2019:
@johnnymatthews @TMoMoreau  Note for future reference: When we create the PR, we should keep in mind that both PRs ( #1067 and #1065 ) address the same issue. I suggest how to coordinate them in #1067.

hsanjuan:
This documentation is generated from https://github.com/ipfs/go-ipfs/blob/master/core/commands/pubsub.go

PR #1069, opened 2022-03-09
single-page-website.md Pinata docs fix

RuniteOre:
The tutorial had old screenshot of Pinata, does not match current layout, updated and updated the steps needed to continue the tutorial without hassle


johnnymatthews (approved):
Oh excellent! That's the downside of taking screenshots I guess, but thanks for updating things for us!

PR #1088, opened 2022-03-21
fix: remove unmaintained copy of go-ipfs config.md

### LINKED ISSUE #1087: Remove or generate /how-to/configure-node
https://docs.ipfs.io/how-to/configure-node/ seems to be based on older version of GO config from  https://github.com/ipfs/go-ipfs/blob/master/docs/config.md. 
The config changes in every version, and there are differences between GO and JS versions.

Right now, a  lot of information on the page is missing, other things are outdated.
For example, `Gateway.PathPrefixes`  were deprecated a while ago (https://github.com/ipfs/go-ipfs/pull/7994) and the page still lists them. 

While this could be fixed manually, I suggest we either remove this duplicate, or  make this page automatically generated from [go-ipfs/docs/config.md](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md) every time a new go-ipfs release ships – just like we already do in https://docs.ipfs.io/reference/http/api/

lidel:
Closes  #1087

I believe it is better to remove the copy of the config for now, and just link to https://github.com/ipfs/go-ipfs/blob/master/docs/config.md directly. 

This way we are not stuck with unmaintained copy.

Alternative is to make this page automatically generated from [go-ipfs/docs/config.md](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md) every time a new go-ipfs release ships, but not sure what is the value added by that (we generate RPC API docs, but here we already have docs, no need to generate/duplicate anything)


filecorgi:
- Check of `docs/how-to/configure-node.md` came back clean!

johnnymatthews (approved):

PR #1125, opened 2022-04-27
Update ipfs-updater.md

mrodriguez3313:
Upgrading version numbers to latest ipfs-update tool version 1.8.0

https://discord.com/channels/806902334369824788/847893063841349652/968664117160341514

With regards to the automation of the docs version number. Maybe there can be a variable in a config file that holds the version number of the tool. With my experience with Jekyll sites, (idk what this one is) but there would be a variable that could look like this `ipfs-update-tool : 1.8.0` and in each location where the version number goes (ex: line 20, etc...) we instead put `{{ site.ipfs-update-tool }}`. Now instead of manually changing each version number in this doc you just change the config file.

### DIFF
--- a/docs/install/ipfs-updater.md
+++ b/docs/install/ipfs-updater.md
@@ -17,22 +17,22 @@ You can download pre-built binaries from [`dist.ipfs.io`](https://dist.ipfs.io/#
 
    ```powershell
    cd ~
-   wget https://dist.ipfs.io/ipfs-update/v1.7.1/ipfs-update_v1.7.1_windows-amd64.zip -Outfile ipfs-update_v1.7.1_windows-amd64.zip
+   wget https://dist.ipfs.io/ipfs-update/v1.8.0/ipfs-update_v1.8.0_windows-amd64.zip -Outfile ipfs-update_v1.8.0_windows-amd64.zip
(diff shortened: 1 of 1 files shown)


johnnymatthews:
Nice thanks! We should probably write a script to do this for us at some point...

johnnymatthews:
>  With my experience with Jekyll sites

Nah, this is a VuePress site. But works in a similar way.

The only problem with having variables like that built into the markdown, is that it makes the markdown itself harder to read. It's also less portable. It'd be better to have a script with a list of files it needs to look at, and then scans those files periodically to check whether they're showing the latest `ipfs-update` version.

Anyway, thanks for this PR. It's in draft mode right now though, are you still working on it?

mrodriguez3313:
The reason I had it as a draft was to see if the conversation of automating it would develop. And if it would be something that I could help take care of? It could be its own seperate PR anyways and I could change this to done.

johnnymatthews (approved):

johnnymatthews:
Yeah it's best if we created a new issue and PR to address this automation problem.

There's an issue here for discussion: https://github.com/ipfs/ipfs-docs/issues/1128

PR #1136, opened 2022-05-05
Remove broken link to canny

2color:
We no longer use canny for feature requests. 

This PR removes the link from the callout.


johnnymatthews (approved):
Looks good! Thanks for the PR!

PR #1160, opened 2022-05-26
update version of ipfs-cluster-ctl

alexeyshurygin:


docs/install/server-infrastructure.md
```diff
@@ -132,7 +132,7 @@ This shows us that `QmdzvHZ...` is pinned across the three IPFS nodes within our
 1. When you're finished playing around, kill the cluster. You may have to run this as root:
 
     ```shell
-    docker-compose kill
+    sudo docker-compose kill
```
Jorropo, line 135: `sudo` here seems weird.

This is up to local setups and your user being in the docker group or not.
There is big security implications of giving it out the docker group, and not having it on your main user probably make sense.

However our docs assume you do, many default installs do that.
It's also very easy for users to figure out what is the issue just by googling the error message if they stamble on it.

Jorropo (changes requested):

lidel:
Thank you @alexeyshurygin  for poking at this. 

We have more cases like this, and they were out of date.  To solve it globally, we've added automation that should bump these version numbers in https://github.com/ipfs/ipfs-docs/pull/1169, and it also bumped ipfs-cluster-ctl so I am closing this one.

PR #1172, opened 2022-06-08
glossary: add CAR

lidel:
cc @olizilla @rvagg  – lmk  if  there are better resources/specs/explainers


filecorgi:

rvagg (approved):
👌 lgtm. If you wanted you could expand slightly on the CARv2 distinction - that v2 is mainly useful for server-side applications and v1 is more useful as a transport; but then again it's really up to users I guess.

PR #1175, opened 2022-06-09
fix: detect new go-ipfs tag using github releases

lidel:
This will check the latest stable release, which will always do the right thing.

## Why old way was buggy?

Old version assumed the latest tag without revision suffix is the latest release but it is not always the case (error prone).

For example, we now have 0.13, but:

```console
$ git clone https://github.com/ipfs/go-ipfs.git
$ cd go-ipfs
$ git describe --tags --abbrev=0
v0.12.2
$ git describe --tags
v0.12.2-506-gb36044be3
```


filecorgi:

PR #1177, opened 2022-06-12
Replaced all 0.12.0 versions with version 0.13.0

Arash-Sabet:


lidel (approved):
Thank you.

Seems we have a bug in our automation, filled https://github.com/ipfs/ipfs-docs/issues/1186

PR #1209, opened 2022-07-21
add storj peer ids

kaloyan-raev:
Peer IDs and addresses of the IPFS nodes of the Storj IPFS pinning service.


johnnymatthews (approved):
Looks good! Tweaked the page a little bit, and alphabetised the list.

PR #1218, opened 2022-07-26
Adds `sudo` to `bash install.sh` command

TMoMoreau:
Found out the other day that when attempting to install on macos, running just `bash install.sh` will move the binary but return an error. Then if you attempt the install again using `sudo` it will return an error of `mv: ./ipfs: no such file or directory` because the first attempt already moved the binary. If this happens, users need to start from the beginning and re-download the binary, this time using `sudo`.


filecorgi:

johnnymatthews (approved):
Nice, thanks!

PR #1254, opened 2022-08-18
[Fixes] Update Hugo instruction

lancatlin:
`relativeURLS` should be `relativeURLs` according to https://gohugo.io/content-management/urls/#relative-urls


TMoMoreau (approved):
Looks good to me, thanks for the help.

PR #1255, opened 2022-08-18
Mention privacy implications of reproviding settings

ghost:
Related to #https://github.com/ipfs/kubo/pull/9207


Winterhuman (commented):
Maybe the wording should be changed to:

Setting the reprovider to "roots" provides the most privacy, it only advertises "direct" or "recursive" pins and not "indirect" pins, however, bare in mind that Bitswap will still respond to requests for indirect pins if the CID is present in the wantlist.

PR #1274, opened 2022-09-06
feat: prox plausible traffic

2color:
Based on https://github.com/ipfs/ipfs-blog/pull/467 which is already working.

This just adds the proxying functionality.

### DIFF
--- a/docs/.vuepress/head.js
+++ b/docs/.vuepress/head.js
@@ -49,9 +49,13 @@ module.exports = [
     }
   ],
   ['meta', { name: 'theme-color', content: '#5bbad5' }],
-  ['script', {
-    defer: true,
-    'data-domain': "docs.ipfs.tech", 
-    src: "https://plausible.io/js/plausible.js"
-  }, ``],
+  [
+    'script',
+    {
+      defer: true,
+      'data-domain': 'docs.ipfs.tech',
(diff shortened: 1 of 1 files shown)


filecorgi:

lidel (changes requested):
sgtm but we need set up `foo.ipfs.tech` for this, so there are no third-party names involved.

Jorropo:
@lidel we can take a look at this again now that ipfs.tech is setup ?

2color:
@Jorropo Based on some tests we did, there's not much benefit to this unless we use the same origin. I can't remember the details, but delegating _only_ a subdomain to Cloudflare is not possible on a normal unless you pay a lot of $.

There may be better ways of doing this using Netlify Edge or Deno Deploy, assuming we can easily do subdomain delegating (which based on recent experiences is easier than Cloudflare)

ElPaisano:
Hey @2color going through open PRs right now. 

Is this still relevant or can we close it out? If it's still relevant, does work still need to be done on this? LMK

2color:
There's not benefit to this with the current domain. It would need to be the same origin to bring benefit.

ElPaisano:
Based on @2colors comment, closing this PR out

PR #1330, opened 2022-11-21
Create pull request link check action and related config file

ElPaisano:
This PR implements https://github.com/tcort/markdown-link-check as a GitHub action. 

**Purpose:** to catch broken links before they get pushed to the live site.

**Files added:**

- `action.yml`: GH Action file that defines when markdown-link-check should run, what version is used, custom variables, config file to be used.
- `mlc_pull_req_config.json`: config file. Currently fixes known markdown-link-check bug where GH links fail for no reason and specifies status codes that should be counted as alive.

**What does this GitHub action do?**

- Runs on pull requests to branch `main`. Specified by `on` and `base-branch` in `action.yml`
- Checks modified files only. Specified by `check-modified-files-only` in `action.yml`
- Shows detailed HTTP status for checked links. Specified by `use-verbose-mode` in `action.yml`
- Handles known bug where GH links return error. Specified in `httpHeaders` in `mlc_pull_req_config.json`
- Valid codes are 200, 206, 429. Specified by `aliveStatusCodes` in `mlc_pull_req_config.json`

**Tests / Input needed**
1. Test: Need to double check that this works as expected with some MD file that hit various link cases. Will specify tests below.
2. Need input from the team:
    - What status codes do we want to count as alive? I currently set 200, 206, 429.
    - Do we cant to add anything else to the config file? Options are here https://github.com/tcort/markdown-link-check#config-file-format. Stuff related to retries, ignorePatterns and replacementPatterns might be useful. 
3. We should consider if we want the checker to ignore certain links. If so, we will need to add HTML comments for said links https://github.com/gaurav-nelson/github-action-markdown-link-check#disable-check-for-some-links in the main branch


ElPaisano:
Actually, I just realized that we probably can't test this before pushing it to `main` with the config I have - this is because I've specified that it should only run on branch `main`. So, need to update this for testing purposes.

johnnymatthews:
Is this workflow compatible with the [docs-starter](https://github.com/protocol/docs-starter)?

johnnymatthews:
> GH MD doesn't allow for relrefs, right?

Correct. Relrefs are strictly a Hugo thing.

johnnymatthews (approved):

PR #1334, opened 2022-11-23
Ci cd test/test markdown link check

TMoMoreau:
My own PR that I'm using to help @ElPaisano test the automated link checker.

Tests that I've run:

1. Curiosity test, changed a URL that was inside a code block, just to see how the link checker worked. It did not pass this test as the URL was not actually a link.
2. Tested on an IPFS gateway link. I changed the CID in the link to a non-existent CID. Passed this test, the checker successfully caught and warned that this link was dead.
3. Tested on a relative link to a different page in the docs. Removed the `.md` from the end of the link. Passed this test, checker successfully found the dead link. Also fixed the IPFS gateway link from the previous test, passed this as well, checker reporting that the link works.
4. Tested header links at varying "depth" (`#`, `###`, `####`, `#####`). Test passed, all links reported as working.
5. Tested broken header links. Link checker is reporting `Test 1` and `Test 2` headers as working, but they're not proper links and don't lead any where. Unsure exactly why this is happening, my guess is that because they don't lead anywhere, it's still technically not a broken link, as in, it returns a 200 OK status.


TMoMoreau:
Tests that I've run:

1. Curiosity test, changed a URL that was inside a code block, just to see how the link checker worked. It did not pass this test as the URL was not actually a link.
2. Tested on an IPFS gateway link. I changed the CID in the link to a non-existent CID. Passed this test, the checker successfully caught and warned that this link was dead.

ElPaisano:
This is great! nice tests. Thanks for doing this @TMoMoreau

TMoMoreau:
Seems like this can be closed now.

PR #1335, opened 2022-11-23
Add Markdownlint Github action 

ElPaisano:
For the initial implementation, tried to configure the rules  to match https://docs.ipfs.tech/community/contribute/grammar-formatting-and-style/#style.

Rule definitions and descriptions are [here](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md)

The rule configuration is defined in `.markdownlint.jsonc`. Each rule has comments that describe whether or not it's already in the style guide / it's something we already enforce without defining it in the style guide / if it might be a good rule to enforce

**Todo:**
- Determine what other configurations the `markdownlint-action.yml `file should have
- Check that the config (`.markdownlint.jsonc`) actually matches https://docs.ipfs.tech/community/contribute/grammar-formatting-and-style/#style
- Run tests for each rule to make sure that the markdownlint check catches violations / doesn't produce false results
- For rules that ARE NOT defined in https://docs.ipfs.tech/community/contribute/grammar-formatting-and-style/#style, determine if we actually want these rules and, if we do, ensure that they won't cause problems for writers

### DIFF
--- a/.github/workflows/markdownlint-action.yml
+++ b/.github/workflows/markdownlint-action.yml
@@ -1,0 +1,14 @@
+name: markdownlint check
+
+on: [pull_request]
+
+jobs:
+  markdownlint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: markdownlint-cli
+        uses: nosborn/github-action-markdown-cli@v3.2.0
+        with:
+          files: .
+          config_file: .markdownlint.jsonc
+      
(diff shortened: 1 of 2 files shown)


ElPaisano:
@johnnymatthews can I get your review on this one? Same idea as the last two linters - enable this as a GH action, but don't require that it be passing to merge things, then iterate on it

johnnymatthews (approved):
Looks fun! Approved.

PR #1338, opened 2022-11-28
cspell gh action intitial commit

ElPaisano:


ElPaisano:
Tabling work on this because Vale has spellcheck plus some. Vale being worked on in this pr https://github.com/ipfs/ipfs-docs/pull/1339

PR #1339, opened 2022-11-28
Initial vale commit

ElPaisano:
Implements [Vale prose linter](https://vale.sh/docs/vale-cli/overview/) as a GitHub action that is triggered on pull requests to ipfs-docs.


ElPaisano:
Updated thoughts on testing:
If we do create a custom IPFS docs-specific (or PLN-wide at a later date) style package, I think it would be a good idea to have test features and fixtures similar to how the MSFT package does it https://github.com/errata-ai/Microsoft#repository-structure. They use [aruba](https://github.com/cucumber/aruba). This is cleaner, quicker and more comprehensive in the long run than my usual hacky way of just making some changes to a md file and pushing them to a separate branch.

ElPaisano:
Ended up opting for a "custom" spelling set up here because the standard way of using a "accept.txt" was not catching everything it should. Will need to continue to experiment with this

johnnymatthews (approved):

PR #1359, opened 2022-12-08
Pull request template

ElPaisano:
Proposing the creation of a PR template similar to how we have templates for Issues.

We can iterate on this over time.

I think it would be helpful for standardization, project management and quality control.

Also, in the initial version, I'm adding a note about the [markdown-link-check GH action](https://github.com/ipfs/ipfs-docs/pull/1330) so that folks are aware it will be in "beta" for awhile after it's initial deployment.


TMoMoreau:
Cool, I agree this is a good idea. I think the fields in there are likely all we would need. It's straightforward and clear what is required, and I think it covers all the bases.

TMoMoreau:
Something to potentially add to this: Add in something near the top to guide people towards opening an issue first, before creating the PR. If the issue is already open, link to the issue in the PR that you are opening.

ElPaisano:
@TMoMoreau , after thinking about this some more, I decided to add the instructions for the beta link checker stuff to https://gist.github.com/ElPaisano/398b4fe36c19c7dcb31cf0220ff43476. I'd like to merge this as is to test its behavior in the GH UI. Once we confirm that it works like it should (i.e. we can click a button to create  a PR and that PR automatically populates with the content in the template), I'm going to add in a blurb like "Please read about the new GitHub link check action at https://gist.github.com/ElPaisano/398b4fe36c19c7dcb31cf0220ff43476".

I elected to initially create the instructions as a gist, so that we don't add them to the docs prematurely (i.e. in case we have to rollback the GH action integration, we decide we don't want to use it, etc.) and the PR template was already pretty full, which I fill like might just confuse people. We can iterate on the instructions and then add them to the docs once the action is out of beta, etc.

ElPaisano:
@TMoMoreau  feel free to comment on https://gist.github.com/ElPaisano/398b4fe36c19c7dcb31cf0220ff43476 with any feedback btw.

If this looks ok otherwise, can we go ahead and merge this?

TMoMoreau (approved):
Looks like a solid template. Merge when ready.

PR #1364, opened 2022-12-12
feat: multiaddresses

### LINKED ISSUE #247: High-level explainer of multiaddresses.
At the Berlin developer summit, we decided *multiaddresses* or *peer addresses* were important to write a concept doc about.

We did not have time to write a rough draft, so we need to create a concept doc from scratch. Concept docs are concise, high-level descriptions of various core concepts in IPFS. They should strive to answer:

1. **What** is this?
2. How does it **relate** to the rest of IPFS?
3. How can (or should?) you **use** it? (Note: sometimes the answer is that you should understand it, but *not* use it. *Bitswap* is probably a good example.)
4. **Where** do you go to learn more?
5. What is the **current state** of affairs?

DannyS03:
## Context

- Provides a short explainer on multiaddresses and a common reference point
- resolves #247

## Latest preview

Please view the latest Fleek preview [here](https://bafybeihptfxmpaeydoyc7n45eo2q37r7azw7orfzi5hc4yt2gfltuueybq.on.fleek.co/concepts/multiaddr/).

### DIFF
--- a/docs/concepts/multiaddr.md
+++ b/docs/concepts/multiaddr.md
@@ -6,6 +6,7 @@ description: Multi-addresses and how they are used in IPFS
 # Understanding Multi-Addresses in IPFS
 
 A [multiaddr](https://github.com/multiformats/multiaddr) encodes multiple layers of addressing information into a single "future-proof" path structure.
+
 It is a self-describing format that allows multiple protocols encoded in a single string. This allows for flexibility in addressing and routing network traffic in a decentralized system like IPFS.
 
 A multiaddr has the following format:
@@ -26,7 +27,7 @@ For example, a multiaddr for an IPFS node with the IP address `192.0.2.1` and li
 
 In IPFS, multiaddrs represent the addresses of nodes on the network. They can be used to specify the source and destination of network traffic, allowing for decentralized routing.
 
-In IPFS, the CID and multihash can be included in a multiaddr to specify the location of a specific piece of content on the network.
+The CID and multihash can be included in a multiaddr to specify the location of a specific piece of content on the network.
 > Each file or piece of content is identified by a unique [content identifier (CID)](content-addressing.md) that is generated based on the content itself. The CID is a base58-encoded string that users can use to retrieve content.
 > In addition to the CID, IPFS uses a multi-hash concept to ensure the content's integrity and authenticity. A multi-hash is a hash of the content generated using multiple hashing algorithms. This allows IPFS to support a variety of hashing algorithms and also provides a level of redundancy to ensure the accuracy of the hash.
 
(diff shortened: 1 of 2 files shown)


filecorgi:

filecorgi:

docs/concepts/multiaddr.md
```diff
@@ -0,0 +1,40 @@
+---
+title: 'Multi-addresses'
+description: Multi-addresses and how they are used in IPFS
+---
+
+# Understanding Multi-Addresses in IPFS
+
+A [multiaddr](https://github.com/multiformats/multiaddr) encodes multiple layers of addressing information into a single "future-proof" path structure.
```
ElPaisano, line 8: Suggestion (with a question):

Since this is the first time "multiaddr" is used on the page, would it make sense to use the term "Multi-address" since that's used in the tile of the doc? Most readers will probably infer that multiaddr is short for multi-address, but I don't think we should assume.  So it would read
```
A [multi-address](https://github.com/multiformats/multiaddr) (multiaddr) encodes multiple layers...
```

Also, is it "multiaddress" or "multi-address"? I'm unsure from https://github.com/multiformats/multiaddr

docs/concepts/multiaddr.md
```diff
@@ -0,0 +1,40 @@
...
+
+It is a self-describing format that allows multiple protocols encoded in a single string. This allows for flexibility in addressing and routing network traffic in a decentralized system like IPFS.
```
ElPaisano, line 10: Suggestion: could consolidate these two points and remove the phrase "future-proof".  Also, I think there is a grammar error here:

```
A multi-address (multiaddr) is a self-describing format that allows multiple protocols to be encoded into a single string. This allows for flexibility in addressing and routing network traffic in a decentralized system like IPFS.
```

docs/concepts/multiaddr.md
```diff
@@ -0,0 +1,40 @@
...
+
+A multiaddr has the following format:
+
+```shell
+/<protocol>/<address>
+```
+
+Where `<protocol>` is the name of the network protocol being used, and `<address>` is the address of the node on that network.
+
+For example, a multiaddr for an IPFS node with the IP address `192.0.2.1` and listening on port `5001` would be:
```
ElPaisano, line 20: You describe the mutliaddr example twice, so I'd suggest deleting the note (`>`) and consolidating

```
For example, this multiaddr indicates that the node uses the IPv4 protocol, has the IP address 192.0.2.1, and listens on TCP port 5001.
```

ElPaisano (changes requested):
Hey @DannyS03 added a few comments

PR #1404, opened 2023-01-10
NOT READY TO MERGE - Proposed Reorganization 1

ElPaisano:
# Don't merge this PR

Proposed reorganization of ipfs-docs. 

[**Latest draft**](https://bafybeib5qq4p33togvqsv5m2cjqtrsinhklpkr4vwwcfwe4nlfogevesrm.on.fleek.co/)


docs/.vuepress/config.js
```diff
@@ -268,18 +249,7 @@ module.exports = {
             ]
           },
           {
-            title: 'Case studies',
```
2color, line 271: I would leave these. We relatively recently moved these up a level and I think this is still a good thing to bring visibility to

Issue #1397, opened 2023-01-06
/api/v0/key/export doesn't exist


TheDiscordian:
Kubo apparently doesn't have an `/api/v0/key/export` endpoint (even though it'd be *incredibly* useful): https://github.com/ipfs/kubo/issues/8327#issuecomment-893450924

Docs entry: https://docs.ipfs.tech/reference/kubo/rpc/#api-v0-key-export

TMoMoreau:
So, should `/api/v0/key/export` be taken out of the docs completely? Or would it be better to leave it in there, but add a note saying that it doesn't have an endpoint and will always fail.

TheDiscordian:
@TMoMoreau I think it should be omitted from the RPC API, as it doesn't appear to be a part of the RPC API.

TheDiscordian:
Maybe we should just omit everything with `NoRemote: true`: https://github.com/ipfs/kubo/blob/e550d9e4761ea394357c413c02ade142c0dea88c/core/commands/keystore.go#L172

TMoMoreau:
@lidel do you have any thoughts about this issue?

hacdias:
@TMoMoreau I think that changing this can be done by adding a `|| cmd.NoRemote` to this line: https://github.com/ipfs/ipfs-docs/blob/main/tools/http-api-docs/endpoints.go#L101. That will ignore all the endpoints that are not available as a remote RPC command.

PR #1348, opened 2022-12-04
Fix broken links, reformat links

ElPaisano:
Wrote a script to run markdown-link-check across the whole repo and found some more broken / improperly formatted links

Files updated:

PR #1423, opened 2023-01-18
Changes to RPC API doc generation to ignore `NoRemote: true` 

TMoMoreau:
# Describe your changes

Change to one line, based on instructions from @hacdias in this comment: https://github.com/ipfs/ipfs-docs/issues/1397#issuecomment-1386861881

I'm not 100% certain that I did this correctly.

# Files changed

- \ipfs-docs\tools/http-api-docs\endpoints.go

- Addresses issue #1397 

No

## Checklist before merging


ElPaisano:
checked out this branch locally and I was also seeing that `/api/v0/key/export` is still listed in the Kubo RPC API doc. 

I tried a few things, in addition to the suggestion https://github.com/ipfs/ipfs-docs/issues/1397#issuecomment-1386861881:
- Seeing what would happen if I set the variable `ignore` in `Endpoints` to be true when `name == "/api/v0/key/export"`:
  `ignore := cmd.Run == nil || IgnoreEndpoints[name] || name == "/api/v0/key/export"`
- Seeing what would happen if I set IgnoreEndpoints
  ```
  var IgnoreEndpoints = map[string]bool{
    "/api/v0/key/export" : true,
  }
  ```
  
  In both cases, didn't fix the issue, but also I'm not super familiar with Go :D
  
   @hacdias any other thoughts on this? Also, are the RPC API pages being autogenerated by scraping info from these pages https://github.com/ipfs/kubo/tree/master/core/commands?

ElPaisano:
@TMoMoreau on a related note, is this the only instance that we are aware of where an RPC command is being autogenerated when it shouldn't be? Or are there other incorrect endpoints in the RPC API doc?

hacdias:
@TMoMoreau the PR is working as intended. However, the CI only updates the docs at a scheduled job once a day if there is a new Kubo version. That is, the Kubo RPC docs are not regenerated at each commit. At least, that is my understanding from the actions you can find under the `.github` directory at the root of the repository. 

I pushed [d03cd36](https://github.com/ipfs/ipfs-docs/pull/1423/commits/d03cd3607c49312d8e6b3aca97df889e19bdf336) to update the README documentation in the generation tool to include the new file path for the RPC docs. To run the tool, you should navigate to `tools/http-api-docs` and run:

```
go run ./http-api-docs/main.go > ../../docs/reference/kubo/rpc.md
```

@ElPaisano I noticed you updated the way the links work in #1348. Unfortunately, if you execute the command above, those nice links will be reverted to the wrong links we used to have before. I pushed [fe969ac](https://github.com/ipfs/ipfs-docs/pull/1423/commits/fe969ac33533c61c139f0499339ba5ab0f02e708) to fix those links of the RPC API at the root, that is, in the generator tool.

---

I run the tool and pushed [d2be7fb](https://github.com/ipfs/ipfs-docs/pull/1423/commits/d2be7fb008b89a725e1d105781bbcec69912759f) with the new Kubo RPC docs. You will see that at least 8 endpoints were removed since they have the `NoRemote` option set to `true`. Here's the preview: https://bafybeidrje25kxpmb5c6kym7g2cjqsgdfkpq7cuzijbxeqf4wpdc6yt7du.on.fleek.co/

hacdias (approved):

hacdias:
@TMoMoreau @ElPaisano I merged it as it was working as intended. Let me know if you have any further questions!

ElPaisano:
@hacdias thanks for the help and the explanation

PR #1448, opened 2023-01-25
Fix typo in nodes

2color:


TMoMoreau (approved):
Good catch. Thanks Daniel.

PR #1452, opened 2023-01-25
Proposed layout adjustment

ElPaisano:
# Preview link 👉 [here](https://bafybeibzjm7huyo42yhyzqgxkbumni736i7u6uw4z2ieaaxv5syewfh75m.on.fleek.co/)

## Edit

I scoped this PR poorly and should have fixed the word bleed issue separately / more quickly. So, broke that work out into https://github.com/ipfs/ipfs-docs/pull/1472, so there's more time to review this, since this PR is more cosmetic preference, and not an actual bug like the css word bleed issue 

## Describe your changes
While there are nice things introduced with the recent layout update for https://docs.ipfs.tech/, we can improve a few things

So, after further thought, while I do like the sidebar - topbar combo on 99% of the site, I don't really feel that the sidebar with the same content as the top menu adds much to the main site page 

image

plus, in mobile, it's just kind of confusing / awkward and takes up unnecessary space 

image

## What this PR does

So, this PR does two things:
- Removes the sidebar from the main page but keeps custom sidebars for all other sections / pages
- Tries to make the site a little more mobile friendly 

## 👋 Request for reviewers

My ask for the reviewer would be to look at [the staging link](https://bafybeih6sa7p5kfn7bddsy3x3fwpe3jlc3mesdhwcv5zj776lxczgrdzxe.on.fleek.co/) and see what the menu looks like in a few screen sizes to confirm that these updates help / don't intro other bugs. Open to any improvements to the css

Right clicking a page in most browsers and clicking **Inspect** will usually open a web console where screen size testing can be done - ping @ElPaisano if you need help figuring out how to do this

# Files changed

## Checklist before merging

### DIFF
--- a/docs/.vuepress/theme/styles/index.styl
+++ b/docs/.vuepress/theme/styles/index.styl
@@ -69,6 +69,10 @@ code {
 	word-break: break-word;
 }
 
+div.legacy-callout section{
+	display: block;
+}
+
 :root {
 	--accent-color: $accentColor;
 	--text-color: $textColor;
--- a/docs/.vuepress/theme/styles/index.styl
+++ b/docs/.vuepress/theme/styles/index.styl
@@ -86,8 +86,19 @@ section.tabs-component-panel::before {
     margin-top: -10em;
 }
 
-.nav-links{
-    margin-right: 10%;
-	margin-left: 10%;
-	float: left;
+@media (min-width: 1060px) {
+	.nav-links{
+		margin-right: 10%;
+		margin-left: 10%;
+		float: left;
+	}
+}
+
+@media (max-width: 720px) {
+	.sidebar-links{
+		visibility: hidden;
+	}
+	.sidebar{
+		width: 30%;
+	}
 }
--- a/docs/.vuepress/theme/styles/index.styl
+++ b/docs/.vuepress/theme/styles/index.styl
@@ -94,11 +94,8 @@ section.tabs-component-panel::before {
 	}
 }
 
-@media (max-width: 720px) {
-	.sidebar-links{
-		visibility: hidden;
-	}
+@media (max-width: 720px) and (min-width: 500px) {
 	.sidebar{
-		width: 30%;
+		width: 40%;
 	}
 }
--- a/docs/.vuepress/theme/styles/index.styl
+++ b/docs/.vuepress/theme/styles/index.styl
@@ -94,7 +94,7 @@ section.tabs-component-panel::before {
 	}
 }
 
-@media (max-width: 720px) and (min-width: 500px) {
+@media (min-width: 500px) and (max-width: 720px) {
 	.sidebar{
 		width: 40%;
 	}
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -345,25 +345,6 @@ module.exports = {
               ]
             }
           ],
-          '/': [
-            '/install/',
-            '/concepts/',
-            '/how-to/',
-            '/reference/',
-            '/project/',
-            {
-              title: 'Case Studies',
-              collapsable: true,
-              children: [
-                ['/case-studies/arbol', 'Arbol'],
-                ['/case-studies/audius', 'Audius'],
-                ['/case-studies/fleek', 'Fleek'],
-                ['/case-studies/likecoin', 'LikeCoin'],
-                ['/case-studies/morpheus', 'Morpheus.Network'],
-                ['/case-studies/snapshot', 'Snapshot']
-              ]
-            },
-          ]
         }
       }
     }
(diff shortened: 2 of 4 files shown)


docs/.vuepress/theme/styles/index.styl
```diff
@@ -86,8 +86,16 @@ section.tabs-component-panel::before {
...
-	float: left;
+@media (min-width: 1060px) {
+	.nav-links{
+		margin-right: 10%;
+		margin-left: 10%;
+		float: left;
+	}
+}
+
+@media (max-width: 720px) and (min-width: 500px) {
```
DannyS03, line 97: Just a nit
```suggestion
@media (min-width: 500px) and (max-width: 720px) {
```

DannyS03 (commented):
Just a nit

docs/.vuepress/theme/styles/index.styl
ElPaisano: @DannyS03 thanks for looking at this 🥳 

Question: as my css is rusty, I'm not clear what this change fixes? Tested it locally, and it appears to be the same behavior. Or is it just to make the css look nicer (`min-width` before `max-width`)?

DannyS03: Ah, right, and yes! Just to make it look nicer (i.e. best practice and being nit-picky)

ElPaisano: Nice makes sense

ElPaisano:
@DannyS03 I think this one is ready to merge, can you approve?

johnnymatthews:
Looks solid from my end. One alternative fix to issue 1 is just to make a smaller IPFS Docs logo. Maybe just use the IPFS cube icon.

Apart from that, this looks good to go!

DannyS03:
> Details

Improving / generating a dedicated landing page from the homepage would be good. I agree that it looks a bit odd with no sidebar (which is mainly because of the current layout and not that the sidebar adds something specifically valuable, as those references can be present on the homepage in other forms.) Happy to help with this.

DannyS03:
> Looks solid from my end. One alternative fix to issue 1 is just to make a smaller IPFS Docs logo. Maybe just use the IPFS cube icon.
> 
> Also, the homepage looks a bit _odd_ with no sidebar:
> 
> 
> It might make more sense to add this page into `get-started` and then set _that_ page as the root of the site. Either that, or justify-centre everything on that homepage. Thoughts? @ElPaisano @DannyS03 @TMoMoreau

> Details

Improving / generating a dedicated landing page from the homepage would be good. I agree that it looks a bit odd with no sidebar (which is mainly because of the current layout and not that the sidebar adds something specifically valuable, as those references can be present on the homepage in other forms.) Happy to help with this.

ElPaisano:
Alright @johnnymatthews @DannyS03 @TMoMoreau 

I agree the main page does look a little odd / empty without the sidebar. So, I:

- widened the content container to take up the remaining space so that it doesn't hopefully look as weird
- widened the footer and wrapped the footer links underneath (they also looked odd)

Still works, looks good in mobile and sidebars look decent on the other pages. 

That being said, because the content is now taking up the whole page, photos should potentially be resized. Might also want to add margins or padding around the content container

RE _It might make more sense to add this page into get-started and then set that page as the root of the site_ - > with the current way we are organizing content, I feel like that might be a little too specific of a category for the landing page

RE _change / resize logo_ _> makes sense, but with this adjustment, I don't think we need to

So I feel like this could get the job done for the moment. 

Thoughts?

image

image


image

ElPaisano:
@johnnymatthews @DannyS03 @TMoMoreau any thoughts on this?

ElPaisano:
@2color can you take a quick look at this PR?

TLDR
1. Widens layout i.e. removes empty space on right/left sides 
2. Removes sidebar ONLY on landing page
3. Fixes a minor css text-bleed issue 

1 and 2 are the bigger changes here, 3 is something I need to do either way with the new layout

2color (changes requested):
> Widens layout i.e. removes empty space on right/left sides

The wider layout diminishes the readability of long sentences on wide screens and doesn't seem to be an established practice.  Moreover, the top bar menu floats to the right which makes it seem less prominent than it should (see screenshot)

Screenshot 2023-02-02 at 2 55 09 PM


For reference, I looked at:
- https://render.com/docs/web-services
- https://www.prisma.io/docs
- https://vercel.com/docs
- https://redis.io/docs/


All of them constrain their main section on wide screens.


> Removes sidebar ONLY on landing page

👍, Seems redundant to have the same options in the top bar and sidebar. However, the typeface for the top menu could pop out a little more; at least as much as the top-level items in the sidebar menu: 

Screenshot 2023-02-02 at 2 56 10 PM

Issue #1465, opened 2023-01-30
js-ipfs docs-wide improvements


ElPaisano:
## Task 1

Anywhere in the site that mentions js-ipfs:

- link to blog post
- warning about js-ipfs being discontinued 
- Add links to helia

Addressed by https://github.com/ipfs/ipfs-docs/pull/1466

## Task 2

Add Helia / links to implementation lists on site

Issue #1459, opened 2023-01-27
Rewrite "Simple chat app" tutorial


2color:
## Background 

The impressive peer-to-peer chat app guide that @TheDiscordian created is no longer functional for several reasons:
- bug in the [Circuit Relay V1 daemon](https://github.com/libp2p/go-libp2p-relay-daemon/issues/18)
- https://github.com/libp2p/js-libp2p-webrtc-star/ is no longer maintained

Both impose severe impediments on browser-to-browser connectivity with libp2p — which, as it stands, is currently unsupported https://github.com/libp2p/js-libp2p/issues/1462.

## What can we do?

Informing readers about the limitation is vital here — let's ensure we set the right expectations. 

May be helpful to link to the [libp2p connectivity page](https://connectivity.libp2p.io/) for this.

2color:
> The guide is useful, but it's not really useful today because the things it talks about don't really work anymore.

My suggestion would be to remove it for now (with a heavy heart because it was such a great example). We still have it on the blog, but by being in the docs, we're setting users up for failure. 

In the mid-term, I'd suggest we follow the work on Helia and CircuitRelayV2 for browser-to-browser connectivity. Once we have some more tooling to do browser-to-browser connectivity maybe we can overhaul the guide.

ElPaisano:
Step 1: Suggestion is to remove this tutorial
Step 2: @TMoMoreau to sync with @TheDiscordian about new tutorial, blog post

ElPaisano:
@TMoMoreau I think that Step 1 (remove the outdated tutorial, and maybe add a redirect somewhere so that it doesn't just 404) should be done as soon as possible, since it's outdated. Is that something you think you can get to today? Then, can work on Step 2 a little more leisurely

ElPaisano:
Suggestion:
- add banner about tutorial being "archived"
- Remove from main site navigation but don't delete
- Redirect away from this page for the moment

ElPaisano:
Update: https://github.com/ipfs/ipfs-docs/pull/1466 should address Part 1 of this, so we can just leave the tutorial as is for the moment.

TMoMoreau:
With step 1 of this issue finished, I added new labels for step 2.

- [x] Add banner/warning to the current tutorial indicating that it is outdated.
- [ ] Create new tutorial that will actually work if someone tries to follow it.

Unsure about the priority of the second step at the moment, this can be discussed here or in a sync.

2color:
[...]

Issue #1360, opened 2022-12-09
Reorg implementations page into 2 separate lists


TMoMoreau:
Plan is to reorg the implementations page into two separate lists, one list at the top for actively maintained implementations, and another list below that to act as a sort of "grab bag" of legacy or deprecated implementations. Currently, all of the implementations are listed together.

https://docs.ipfs.tech/basics/ipfs-implementations/

TMoMoreau:
What are we considering "active" when it comes to these implementations?

@2color would you consider Auspinner still "active"?

2color:
I wouldn't really consider it active. But it still works.

I think maybe setting the criteria of activity in the last 6 months might be a good bar for this

ElPaisano:
@TMoMoreau can you link the PR for this?

TMoMoreau:
No PR yet, will link it when it's up. PR will be up shortly.

TMoMoreau:
@ElPaisano PR #1456

PR #1466, opened 2023-01-30
Add warning about js-ipfs being discontinued / point to Helia in a few places

ElPaisano:
# Describe your changes
Addresses https://github.com/ipfs/ipfs-docs/issues/1465 and [Part 1 of issue 1459](https://github.com/ipfs/ipfs-docs/issues/1459#issuecomment-1409131076)

## Checklist before merging

### DIFF
--- a/docs/concepts/ipfs-implementations.md
+++ b/docs/concepts/ipfs-implementations.md
@@ -5,6 +5,13 @@ description: "There isn't just one canonical IPFS implementation. Learn about th
 
 # IPFS Implementations
 
+:::warning js-ipfs being discontinued
+Development of the [js-ipfs project](https://github.com/ipfs/js-ipfs) is being discontinued to focus on [Helia](https://github.com/ipfs/helia), a leaner, more modular, modern implementation of IPFS in JavaScript slated for release in 2023. To learn more about Helia and the current state of IPFS in JS, see the [blog post](https://blog.ipfs.tech/state-of-ipfs-in-js/). 
+
+Because of this, js-ipfs tutorials may be out of date, and will eventually be archived.
+
+:::
+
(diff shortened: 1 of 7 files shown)


ElPaisano:
@2color I added this callout in the following places:

Install IPFS for JavaScript (js-ipfs)
IPFS Implementations
Create a simple chat app
IPFS repository guide > Protocol implementations

Any other spots you think we ought to add this in?

docs/concepts/ipfs-implementations.md
```diff
@@ -5,6 +5,13 @@ description: "There isn't just one canonical IPFS implementation. Learn about th
 
 # IPFS Implementations
 
+:::warning js-ipfs being discontinued
+Development of the [js-ipfs project](https://github.com/ipfs/js-ipfs) is being discontinued to focus on [Helia](https://github.com/ipfs/helia), a leaner, more modular, modern implementation of IPFS in JavaScript slated for release in 2023. To learn more about Helia and the current state of IPFS in JS, see the [blog post](https://blog.ipfs.tech/state-of-ipfs-in-js/).
```
2color, line 9: I would move this either to below the table or into the js-ipfs row so it's not the first thing you read on the page.

Also, this is the first time I've seen `slated` used this way. Looking it up reveals that it's a north American word which probably explains why I haven't seen it. What do you think about a more neutral verb, e.g. scheduled or planned?

2color (changes requested):
The new warning element itself is great.

I would also add it to the following pages:
- https://docs.ipfs.tech/reference/js/api/
- https://docs.ipfs.tech/concepts/nodes/#js-ipfs

2color (approved):

2color:
Either here or in a separate PR, e.g. #1360, we should add Helia to the list of implementation

ElPaisano:
> Either here or in a separate PR, e.g. #1360, we should add Helia to the list of implementation



> Either here or in a separate PR, e.g. #1360, we should add Helia to the list of implementation

👍 made a note to address it as part of https://github.com/ipfs/ipfs-docs/issues/1465, will create a quick PR

PR #1467, opened 2023-01-31
Add CID conversion example with embed

2color:
This adds an embed of Codesandbox to the CID concepts

### DIFF
--- a/docs/concepts/content-addressing.md
+++ b/docs/concepts/content-addressing.md
@@ -69,13 +69,24 @@ $ ipfs cid format -v 1 -b base32 QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR
 bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
 ```
 
-JavaScript users can also leverage the `toV1()` method provided by the [`cids`](https://www.npmjs.com/package/cids) library:
-```js
-const CID = require('cids')
-new CID('QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR').toV1().toString('base32')
-// → bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
-```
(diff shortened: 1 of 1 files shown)


docs/concepts/content-addressing.md
```diff
@@ -60,35 +60,40 @@ Check out ProtoSchool's [Anatomy of a CID](https://proto.school/anatomy-of-a-cid
...
-```js
-const CID = require('cids')
-new CID('QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR').toV1().toString('base32')
-// → bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
-```
+### Programatically with TypeScript/JavaScript
+
+The [`multiformats` npm package](https://www.npmjs.com/package/multiformats) can be used to convert between CID versions, base representations, and access the binary representation:
+
+<iframe src="https://codesandbox.io/embed/working-with-cids-xrvqop?autoresize=1&fontsize=14&hidenavigation=1&theme=dark"
```
lidel, line 76: Including dependency on external service that may dissapear at any time is a bit brittle :-1:

lidel (changes requested):

docs/concepts/content-addressing.md
2color: Are you referring to https://codesandbox.io/? It's a pretty mature service so I think it may be worth considering the benefit vs cost of this.

Do you think that having an interactive/runnable example is useful? I'd argue that yes (even though this page may not be the right one for it and it should go into the guides)

2color: Summary of in-person discussion with @lidel:
The main problem with embedding codesandbox is that it won't work when the page is loaded locally without an internet connection.

While that's true, I think the benefits of having these interactive examples outweigh the cost of them not working without internet connection.

ElPaisano:
@2color @lidel following up on this PR. Is this now addressed by merged PR https://github.com/ipfs/ipfs-docs/pull/1476?
If so, could we close this? If not, lmk if you need content or a review or anything from me to help get it out

PR #1482, opened 2023-02-08
windows installation zip file naming issue fixed

nazmulshuvo03:
# Describe your changes

The zip file downloaded has 'windows-amd64' in its name, but the name is wrong in unzip command

# Files changed

docs/install/command-line.md

I have not searched for any issue related to this. Just found it while using the page

No

## Checklist before merging


TMoMoreau:
I just tested these steps on Windows, and the current steps are correct. After running the `wget` command, you will see that the `.zip` file that was downloaded is named: `kubo_v0.18.1.zip`, which matches the command in the following step. Changing the name of the file in the command would break the command and it would no longer work.

We appreciate you taking an interest in helping with the documentation, but this PR will now be closed without merging.

PR #1484, opened 2023-02-09
More redirects / file shuffling

ElPaisano:
Moves `basics\go\command-line` to `how-to\kubo-basic-cli` and adds redirects

@TMoMoreau please manually test the new redirects in `redirects` using the latest Fleek build link to ensure that we didn't miss anything here


TMoMoreau (approved):
Manually checked the new redirects, everything seems to be good.

PR #1489, opened 2023-02-14
Removes Algolia from docs.

johnnymatthews:
This is a hotfix PR that reverts the site back to using the default VuePress search. Algolia wasn't indexing the site properly, for some reason. We'll merge this PR while the team figures out what's going on with Algolia.

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -26,10 +26,6 @@ module.exports = {
     }
   },
   themeConfig: {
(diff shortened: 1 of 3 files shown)


2color (approved):

lidel:
@johnnymatthews @2color  Could it be because they tried to use `.io` and we force-redirect to `.tech` now? (https://github.com/protocol/bifrost-infra/issues/2018)

PR #1516, opened 2023-03-07
Landing page revamp

ElPaisano:
Created this PR as a starting point for discussion regarding how we can refactor the landing page, which has come up multiple times.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -30,7 +30,7 @@ Are you developing with IPFS implementations and tools, and looking for technica
 
 - Install and set up various IPFS implementations and tools in [Get Started](./install/README.md).
 - Get inspired with [IPFS usage ideas and examples](#examples-and-case-studies).
-- Find reference material for IPFS implementations, tools and the HTTP API in the [API and CLI references](./reference/reference/README.md).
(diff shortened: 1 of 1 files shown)


ElPaisano:
@2color I know that this has been discussed multiple times and got lost in the recent shuffling of priorities. A few comments here:

- The goal is to simplify this page and make it useful for users landing on the site / navigating
- I removed some of the definition stuff at the top, as I feel like this is now covered by pointing the user tohttps://docs.ipfs.tech/concepts/#learn-the-basics 
- Broke it into 3 subsections **Learn**, **Develop** and **Join the community**
- Added two callouts, one for new IPFS-ians to read through https://docs.ipfs.tech/concepts/#learn-the-basics and another for devs to follow https://docs.ipfs.tech/community/#get-technical-support-and-help to get support

This is just the starting point though. Other things:
- We could probably add a little blurb / subsection to point users to the Quickstart guides you've been working on, for users who are just trying to publish content, etc.
- Be a little more prescriptive in the the Develop subsection. Part of the issue here is that developer reference material is spread out across Guides and Reference, so this may not be possible with our current docs set up

Anyways, we can just keep it simple, make a few changes and then iterate later.

2color (approved):
This is great!

PR #1535, opened 2023-03-23
edit to diagnostic tools

ElPaisano:
@2color this removes the reference to IPFS gateway checker tool due to the unresolved warning banner that pops up every time you go to the site. My thoughts are that we shouldn't be linking users to a site that does this with no explanation. Or, we should at leas texplain this behavior to user so they aren't confused - seems like an option for this PR as well


ElPaisano:
Addressed in https://github.com/ipfs/ipfs-docs/pull/1545

PR #1549, opened 2023-04-05
Adds Swiftype, but VuePress search still enabled.

johnnymatthews:
Disables VuePress search and integrates Swiftype.

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -484,7 +484,7 @@ module.exports = {
     [
       '@vuepress/search',
       {
-        searchMaxSuggestions: 10
+        searchMaxSuggestions: 1
       }
     ],
     'vuepress-plugin-chunkload-redirect',
--- a/docs/.vuepress/head.js
+++ b/docs/.vuepress/head.js
@@ -53,4 +53,11 @@ module.exports = [
     'data-domain': "docs.ipfs.tech", 
     src: "https://plausible.io/js/plausible.js"
   }, ``],
+  ['script', {}, `
+    (function(w,d,t,u,n,s,e){w['SwiftypeObject']=n;w[n]=w[n]||function(){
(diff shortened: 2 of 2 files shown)


johnnymatthews:
Ok, this is a pretty hacky PR, so bare with me while I explain what I did. To implement Swiftype, all we have to do is:

## 1. Add a custom script tag to the header

Simple enough. [See `docs/.vuepress/head.js`](https://github.com/ipfs/ipfs-docs/pull/1549/files#diff-42eb0560ebb841f0b56650f08e0b7249ca59d2ecb2838ed534fa9697de0eac3b

## 2. Add a class to the search bar

This _should_ be easy, but VuePress handles the search bar in a strange way. In VuePress, the search bar isn't contained within the built in theme, it's an external package: [`@vuepress/plugin-search`](https://www.npmjs.com/package/@vuepress/plugin-search). That plugin accepts very limited customization options. One modification the plugin does not accept is custom classes.

So, I had to grab the package from `/node_modules/@vuepress/plugin-search`, move it to `/docs/.vuepress/plugins`, and then install it in [`/docs/.vuepress/config.js`](https://github.com/ipfs/ipfs-docs/pull/1549/files#diff-52eb86351480925d61947a58caca1ea76d10950ec26533f16f37cb12a3519dff). I then had to remove the original @vuepress/plugin-search dependency.

Finally, I removed logic from the plugin that shows suggestions -- this is handled by Swiftype already.

## Summary

This is a very hacky way of getting Swiftype to work. Unfortunately it's the only option I can think of right now since VuePress isn't particularly customizable. I also wouldn't expect much help from the VuePress devs since the project has gone into maintenance mode.

ElPaisano:
@johnnymatthews going through and cleaning up old issues and PRs. 

What are your thoughts on this at this point? Worth continuing?

PR #1561, opened 2023-04-13
Add content provider warning / add EEC

ElPaisano:
Stages https://github.com/ipfs/ipfs-docs/pull/1546 because IMO, 1546 indicates that we should add a disclaimer, so this PR add a disclaimer to the content provider list before adding 1546.


johnnymatthews (approved):
Looks fine to me.

PR #1565, opened 2023-04-16
docs: _redirects specification link

SgtPooki:
https://github.com/ipfs/specs/blob/main/http-gateways/REDIRECTS_FILE.md points to https://specs.ipfs.tech/http-gateways/web-redirects-file/ now


ElPaisano (approved):
LGTM

PR #1567, opened 2023-04-17
Remove Implementation Status page

### LINKED ISSUE #1564: Remove outdated implementation status page
The [project/implementation-status page](https://docs.ipfs.tech/project/implementation-status/) uses the outdated dichotomy of go vs. js impls. Two possibilities:

- Delete this page altogether
- Update and move it to Reference section. The endpoints table seems like a useful structure. If we want to keep that (and update for kubo & other impls), it should probably move to the Reference section.

Thoughts @BigLep @lidel @autonome @ElPaisano?

mishmosh:
# Describe your changes
This PR removes [project/implementation-status page](https://docs.ipfs.tech/project/implementation-status/), for a few reasons discussed in #1564:
- It uses the outdated dichotomy of go vs. js impls.
- Instead, we want to have more specs at https://specs.ipfs.tech and a table that shows which spec is implemented by what (but we need to have more specs in presentable state first).
- Manually maintained tables are likely to go stale. Things like this should be driven from test suite results.

# Files changed
1. docs/.vuepress/config.js
2. docs/project/README.md
3. project/implementation-status.md

Fixes #1564.

No

## Checklist before merging
- [x] Passing all required checks (The beta **Check Markdown links for modified files** check is not required)


ElPaisano (approved):
Fixed the build, it was a quirk of the way config.js defines our menus

LGTM to me based on the issue / PR description 👍

ElPaisano:
@mishmosh @BigLep created https://github.com/ipfs/ipfs-docs/issues/1569 based on this PR / issue

ElPaisano:
I think we can ship this as is, unless @lidel do you have any additional feedback?

ElPaisano:
Merging since we have one approval and haven't heard otherwise. Can incorporate additional feedback in a followup PR if need be

BigLep (approved):

BigLep:
Agreed, @ElPaisano , that this can merge.  

Generally I think we should have a "bias for action" posture with docs.  Removing incorrect old information is better than keeping it around and waiting for the ideal replacement.  

Worst case if we get something wrong with docs, we can revert it.  Given the good judgement of team members I don't think we'll have any catastrophic slip-ups.

PR #1613, opened 2023-06-02
Updated Kubo troubleshooting to not name squat

BigLep:
This is an appropriate fallout from the work to renaming go-ipfs to Kubo.


ElPaisano:
LGTM

Issue #1602, opened 2023-05-25
Kubo Install - can we just recommend folks on mac to use homebrew?


momack2:
https://formulae.brew.sh/formula/ipfs seems like an easy way to install, but the docs recommend using curl and then unzipping a tarball. Any reason not to give the easier option?

PR #1655, opened 2023-08-07
Kubo install: update Mac instructions

ElPaisano:
Addresses https://github.com/ipfs/ipfs-docs/issues/1602


ElPaisano:
@lidel this PR updates Kubo install instructions for Mac to recommend https://formulae.brew.sh/formula/ipfs over the current mac install process which uses curl https://docs.ipfs.tech/install/command-line/#install-official-binary-distributions

Could you or someone from the kubo team do a quick review of the updated instructions? https://bafybeiczcyidxtimhmot5rgfhfsrxayt2tas6y6mkcp4c7n5ti3c3kal3q.on.fleek.co/install/command-line/#install-official-binary-distributions 

One thing I'm wondering: would there be any reason to keep the old mac install instructions somewhere in the docs? Or can we just cut them completely and suggest brew?

docs/install/command-line.md
```diff
@@ -147,66 +147,16 @@ For installation instructions for your operating system, select the appropriate
...
-1. Navigate to the `kubo` directory:
-
-   ```bash
-   cd kubo
-   ```
-
-1. Run the install script:
+::: 
+This installation method supports both Intel and Apple Silicon hardware.
+:::
```
lidel, line 152: We should have a short sentence in case someone wants to do manual install or compare artifacts:
(the `v0.21.0` will get updates automatically via automation around `current-ipfs-version`): 

```suggestion
The `brew` installation method supports both Intel and Apple Silicon hardware.

If you prefer a manuall installation, `darwin-amd64` (Intel) and `darwin-arm64` (Apple Silicon) artifacts are provided at https://dist.ipfs.tech/kubo/v0.21.0/

:::
```

lidel (commented):
Overall, sounds sensible, with big caveat below + small suggestion inline.

### On macOS notarization and who do we trust
 
I am not macOS user myself, but my understanding of macOS security is that it is based on reputation of the entity that builds artifact and performs macOS notarization.

If you do something shady, your notarization key is revoked, and all apps that were notarized with it are revoked from running as well.

This PR replaces dist.ipfs.tech artifacts notarized using Protocol Labs keys and infrastructure, with  [brew package](https://formulae.brew.sh/formula/ipfs) that follows [this formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/ipfs.rb) which builds artifacts on their own infrastructure and then publishes binaries at `https://ghcr.io` URLs visible in https://formulae.brew.sh/api/formula/ipfs.json.  I imagine at the end of the build but before publishing, the artifacts are notarized somehow with Brew keys?

### Should we trust and endorse brew?

So the question is: do we trust Brew CI and Github less that we trust Github CI (already used for building dist.ipfs.tech) and our own self-hosting at dist.ipfs.tech? 

To me, it does not sound like we weaken supply chain security for macOS much, if anything, there are more eyes looking at Brew notarized artifacts than currently looking at dist.ipfs.tech.  Pointing at brew may be net positive, as it provides users with a mechanism for getting future updates via brew CLI.

Once again,  I am not macOS user,  would be good to get :+1:  from someone more familiar with Kubo and `brew` 
(cc @hacdias @alanshaw @olizilla  ?)

hacdias:
I personally do think we should list Homebrew. I use it quite a lot on macOS and I even used to have Kubo installed via `brew`. I no longer do it simply because I work on its codebase, so I just need to `make install`. But if I had to recommend someone with macOS, I would definitely just point them at `brew`.

docs/install/command-line.md
```diff
@@ -147,66 +147,19 @@ For installation instructions for your operating system, select the appropriate
...
-1. Decide which macOS binary to download from [`dist.ipfs.tech`](https://dist.ipfs.tech/#kubo) by determining whether your system uses an Apple or Intel CPU. On most macOS systems, you can find systems specs by doing the following:
+::: 
+The `brew` installation method supports both Intel and Apple Silicon hardware.
 
-   1. In the upper left hand corner of your screen, click the "Apple" icon.
-   1. In the drop-down menu displayed, select _About this Mac_.
-   1. A window with information about your Mac is displayed.
-      - If your system uses Apple Silicon, the specific chip displays, such as _Apple M1 Pro_.
-      - If your system uses an Intel CPU, the specific processor displays, such as _2.3 GHz 8-Core Intel Core i9_.
+If you prefer a manuall installation, `darwin-amd64` (Intel) and `darwin-arm64` (Apple Silicon) artifacts are provided at https://dist.ipfs.tech/kubo/v0.21.0/
```
ElPaisano, line 153:
```suggestion
If you prefer a manual installation, `darwin-amd64` (Intel) and `darwin-arm64` (Apple Silicon) artifacts are available at https://dist.ipfs.tech/kubo/v0.21.0/
```

docs/install/command-line.md
```diff
@@ -389,7 +342,7 @@ When an IPFS command executes without parameters, the CLI client checks whether
 
 - If an `$IPFS_PATH` is in the default location (for example, `~/.ipfs` on Linux), then it works automatically and the IPFS CLI client talks to the locally running `ipfs` daemon without any extra configuration.
 
-- If an `$IPFS_PATH` isn't in the default location, use the `--api <rpc-api-addr>` command-line argument. Alternatively, you can set the environment variable to `IPFS_PATH`. `IPFS_PATH` will point to a directory with the api file with the existing `ipfs` daemon instance.
+- If an `$IPFS_PATH` isn't in the default location, use the `--api <rpc-api-addr>` command-line argument. Alternatively, you can set the environment variable to `IPFS_PATH`. `IPFS_PATH` will point to a directory with the `$IPFS_PATH/api` file pointing at Kubo RPC of the existing `ipfs` daemon instance.
```
ElPaisano, line 345:
```suggestion
- If an `$IPFS_PATH` isn't in the default location, use the `--api <rpc-api-addr>` command-line argument. Alternatively, you can set the environment variable to `IPFS_PATH`. `IPFS_PATH` will point to a directory with the `$IPFS_PATH/api` file pointing at the Kubo RPC of the existing `ipfs` daemon instance.
```

ElPaisano (commented):
Minor writing things

PR #1670, opened 2023-08-17
fix typo: ipfs-gateway.md

Miigon:
HTTP-incompatible -> IPFS-incompatible

# Describe your changes
"HTTP-incompatible browsers, tools and software" doesn't make sense. Assuming it's a typo.

# Files changed
- docs/concepts/ipfs-gateway.md

None

None

## Checklist before merging

### DIFF
--- a/docs/concepts/ipfs-gateway.md
+++ b/docs/concepts/ipfs-gateway.md
@@ -11,7 +11,7 @@ related:
 
 # IPFS Gateway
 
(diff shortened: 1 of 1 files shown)


docs/concepts/ipfs-gateway.md
```diff
@@ -211,4 +211,4 @@ No. The ipfs.io gateway is one of many portals used to view content stored by th
 
 - [A Practical Explainer for IPFS Gateways – Part 1](https://blog.ipfs.tech/2022-06-09-practical-explainer-ipfs-gateways-1/), [Part 2](https://blog.ipfs.tech/2022-06-30-practical-explainer-ipfs-gateways-2/)
 - [Kubo: Gateway configuration options](https://github.com/ipfs/kubo/blob/master/docs/config.md#gateway)
-- [IPFS HTTP Gateways specification](https://specs.ipfs.tech/http-gateways/) 
\ No newline at end of file
+- [IPFS HTTP Gateways specification](https://specs.ipfs.tech/http-gateways/)
```
Miigon, line 214: I have no idea why there is a change with identical content before and after. But I've tried both editing with vim locally and editing directly through GitHub (all without touching this line at all) and this change occurs in both case.

My guess is some sort of CRLF mismatch that git tried to correct by itself?

Miigon, line 214: I have no idea why there is a change with identical content before and after. But I've tried both editing with vim locally and editing directly through GitHub (all without touching this line at all) and this change occurs in both case.

My guess is some sort of CRLF mismatch that git tried to correct by itself?

PR #1772, opened 2023-11-30
fix: kubo v0.24.0

lidel:
Seems that manually updating `tools/http-api-docs/go.mod` in https://github.com/ipfs/ipfs-docs/pull/1755 made [.github/workflows/update-on-new-ipfs-tag.yml](https://github.com/ipfs/ipfs-docs/blob/main/.github/workflows/update-on-new-ipfs-tag.yml) skip some work, as it is using `go.mod`  for deciding if docs should be updated to the latest kubo. 

I've run generators manually to fix that. Future 0.25.0  should work fine, as it did for 0.23.0 (https://github.com/ipfs/ipfs-docs/pull/1730)


hacdias:
@lidel I don't have permissions to merge, and `fleek/build` is required. Tried ton trigger it, but it doesn't run

hacdias:
Superseded by #1783

PR #1814, opened 2024-02-16
feat: add public utilities page

2color:
## What this PR does

This PR adds a new page on public utilities infrastructure.

There's some nuance with the wording of who's operating the infra we need to get right. Would be good to hammer this out in this PR

## TODO

- [x] Update gateways page FAQ based on whatever we decide here
- [x] Add information on the Amino DHT bootstrappers (or is that libp2p teritory)

### DIFF
--- a/.github/styles/pln-ignore.txt
+++ b/.github/styles/pln-ignore.txt
@@ -14,6 +14,9 @@ Audius
 auspinner
 BitSwap
 bitswap
+bootstrappers
+bootstrapper
+bootstrap
 Blockchain
 blockchain
 blockchains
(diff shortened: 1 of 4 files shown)


mishmosh:
Apologies for the delay. I made 2 separate commits to (1) update the gateways page especially its TOC and (2) clarify Waterworks/Foundation role in offering public utilities.

lidel (approved):
Lgtm, small suggestions inline.

2color:
I moved the FAQ which had mostly questions about the gateway from the general gateway page here and changed from PL to the IPFS Foundation. I also added the bootstrappers. 

@mishmosh Are the bootstrap nodes also provided by the IPFS Foundation?

PR #1824, opened 2024-03-05
docs: mark inactive kubo rpc api clients

hacdias:
Just marking some Kubo RPC clients as inactive. These clients have had no activity for over 3 years (some 5!).


hacdias:
Oh, I just saw #1823

lidel (approved):

PR #1825, opened 2024-03-05
feat: mention verified-fetch

2color:
# Description

Add link to verified fetch

### DIFF
--- a/docs/reference/js/api.md
+++ b/docs/reference/js/api.md
@@ -8,7 +8,7 @@ description: Developer resources for working in JavaScript with IPFS, the InterP
 Developers can get started with IPFS in JavaScript (JS)  using several options:
 
 - [IPFS Helia](https://helia.io): a lean, modular, and modern implementation of IPFS for the JS and browser environment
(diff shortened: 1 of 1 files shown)


docs/reference/js/api.md
```diff
@@ -8,6 +8,7 @@ description: Developer resources for working in JavaScript with IPFS, the InterP
 Developers can get started with IPFS in JavaScript (JS)  using several options:
 
 - [IPFS Helia](https://helia.io): a lean, modular, and modern implementation of IPFS for the JS and browser environment
+- [@helia/verified-fetch](https://github.com/ipfs/helia-verified-fetch/tree/main/packages/verified-fetch): A [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)-like API for obtaining verified & trustless IPFS CIDs on the web
```
achingbrain, line 11: A slight nitpit but we don't obtain CIDs, the user has the CID already.  Maybe something like this?

```suggestion
- [@helia/verified-fetch](https://github.com/ipfs/helia-verified-fetch/tree/main/packages/verified-fetch): A [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)-like API for obtaining verified & trustless IPFS content on the web
```

achingbrain (approved):

PR #1842, opened 2024-04-13
Add ExIPFS - an active rpc client

bahner:
The referenced library is the core client.

Obviously there is nothing to test with this very simple commit.


2color (approved):
Nice! Thanks

PR #1865, opened 2024-05-16
From August 14th, 2024, Cloudflare hostnames will no longer connect to IPFS

LoueeD:
# From August 14th, 2024, Cloudflare hostnames will no longer connect to IPFS

Updated the documentation to inform users about the upcoming changes to Cloudflare IPFS gateways. Specifically, Cloudflare hostnames will no longer connect to IPFS, and users need to switch to the new hostnames to avoid service interruptions.


2color (approved):

bumblefudge (approved):

lidel:
Thank you from raising this.

This is tricky: IPFS docs are not responsible for third-party services, and adding the whole paragraph here is nto something we did for other decommissioned gateways over the years.

Instead, we should remove mentions of decommissioned gateway host names.
Let's continue in https://github.com/ipfs/ipfs-docs/pull/1865

PR #1910, opened 2024-08-16
Add IPFS "Mainnet" to the glossary 

2color:
# In this PR 
- **feat: add Mainnet to the glossary**
- **feat: add simpler explanation for CAR**

- IPFS Mainnet is a useful meme evident by the organic adoption of the term
- It's very useful for distinguishing between Iroh, experimental implementations and "Mainnet"

## Checklist before merging

### DIFF
--- a/docs/concepts/glossary.md
+++ b/docs/concepts/glossary.md
@@ -76,7 +76,7 @@ A Bootstrap Node is a trusted peer on the IPFS network through which an IPFS nod
 
 ### CAR
 
-The CAR (Content Addressable aRchives) is a format for serialized representation of any [IPLD](#ipld) [DAG](#dag). Typically in a file with a .car filename extension or a byte stream marked as [application/vnd.ipld.car](https://www.iana.org/assignments/media-types/application/vnd.ipld.car) media type. [More about CAR](https://ipld.io/specs/transport/car/)
+The CAR (Content Addressable aRchives) is a format for serialized representation of any [IPLD](#ipld) [DAG](#dag). The CAR format is a way of packaging up content-addressed data into archive files that can be easily stored and transferred. Typically in a file with a `.car` filename extension or a byte stream marked as [application/vnd.ipld.car](https://www.iana.org/assignments/media-types/application/vnd.ipld.car) media type. [More about CAR](https://ipld.io/specs/transport/car/)
 
 ### CAR v1
 
--- a/docs/concepts/glossary.md
+++ b/docs/concepts/glossary.md
@@ -208,7 +208,7 @@ An experimental data store used when `--nocopy` is passed to `ipfs add`. It stor
 
 ### Gateway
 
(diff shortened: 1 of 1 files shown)


docs/concepts/glossary.md
```diff
@@ -298,6 +298,10 @@ In IPFS and [IPLD](#ipld), a _link_ usually means a pointer to some [CID](#cid).
 
 ## M
 
+### IPFS Mainnet
+
+IPFS Mainnet is a term used to describe the default or "main" network that many IPFS implementations connect to. Mainnet is a subset of all the [IPFS implementations](https://specs.ipfs.tech/architecture/principles/#ipfs-implementation-requirements) and typically includes implementations that connect to the [Amino DHT](#amino) and support [Bitswap](#bitswap), [UnixFS](#unixfs), and the [IPFS Gateway](#gateway). This has mostly been assumed for the IPFS network and is not emphasized for anyone other than protocol experts. Nonetheless, IPFS Mainnet is a useful distinction in a world of many [IPFS implementations](../concepts/implementations.md) with varying degrees of interoperability.
```
bumblefudge, line 303: For me, "mainnet" is less about implementations than configs, i.e. a private-network kubo is not on mainnet, but a default-config kubo puts everything it `ipfs add`s onto mainnet, if you catch my drift.  I could be wrong, tho!

```suggestion
IPFS Mainnet is a term used to describe the default or "main" network that most IPFS implementations connect to by default configuration.
Most [IPFS implementations](https://specs.ipfs.tech/architecture/principles/#ipfs-implementation-requirements) were designed to work with mainnet, but many can be configured instead to use private networks (i.e. only sync with manually allowlisted peers) or only serve content generated locally.  

Mainnet as a network and data substrate refers to the [Amino DHT](#amino). comprised of nodes that synchronize their contents using the [Bitswap](#bitswap) sync protocol;
as all data is, by default, converted to [UnixFS](#unixfs) at ingress by these implementations, and [IPFS Gateways](#gateway) are configured to assume [UnixFS](#unixfs)-only data, other encodings are sometimes considered incompatible with or problematic for use with bitswap, and as such, not "Mainnet-compatible".
Most IPFS documentation is written assuming all Mainnet-compatible configurations and network topology, because straying from these defaults can break many assumptions of the system and is recommended only for advanced applications.
For this reason, compatibility with IPFS Mainnet is often referred to as "full IPFS compatibility", and many [IPFS implementations](../concepts/implementations.md) require additional care to interoperate at the bitswap-sync, DHT-advertising, or CID-fetching layers.
```

(totally just winging it, feel free to take or leave any of that, but this would be my "interop maxi"/"roast my docs" take on the full meaning of Mainnet!)

bumblefudge (approved):
Approving because as-is it's better than status quo, but also left some additional content if you have appetite for further wordsmithing and politicizing 😄

bumblefudge:
also I feel like @bmann deserves coining credit at least in the git-trail, if not on the actual docs website?  Or we could add a little introductory clause like, `First introduce by Boris Mann on the [Fission blog](https://fission.codes/blog/ipfs-mainnet-everywhere-computer/),"...

docs/concepts/glossary.md
2color: I've integrated some of the information from your suggestion. For consistency with other glossary terms, I'm keeping this brief. If you feel it's necessary to add all the additional information, I would create a new page dedicated to this topic. But I'd consider that out of scope for this PR.

PR #1923, opened 2024-09-05
feat: add ipfs check video

2color:
## Changes in this PR

Add IPFS Check debugging guide video


2color:
The only check that fails is the markdown link checker. 

I haven't figured out to configure it to ignore YouTube embeds, so merging despite failing check

lidel (approved):

PR #1932, opened 2024-10-01
chore: brave cleanup + updated links

lidel:
Removed mentions of "Brave support" and updated some outdated content I noticed.

Context: https://blog.ipfs.tech/2024-brave-migration-guide/ cc @2color


lidel:
I'm merging this because we've improved HTTP 504 error pages oon public gateways and users will see them more often (30s timeout). We do not want them to hit outdated content and waste time installing Brave.

Fee free to improve in follow-up PR :pray:

PR #1935, opened 2024-10-02
chore: remove content related to legacy `ipfs-ds-convert` tool

lidel:
The tool is being sunset: https://github.com/ipfs/ipfs-ds-convert/issues/50 and the "default-profile" page here was focused on using the tool to convert form flatfs to now deprecated badgerv1. Not something we want users to do.

There is a lot of things in the document that may be a footgun.

For example:  
- "Find your current profile"  suggestes that "datastore" == "profile"
- suggesting `lowpower` profile (know footgun: https://github.com/ipfs/kubo/pull/10524) causing people to waste time, get frustrated with IPFS' content routing "not working" 
- outdated copy of content https://github.com/ipfs/kubo/blob/master/docs/config.md#profiles

 The `ipfs-ds-convert` tool itself was not maintained since 2021 and go-ipfs v0.8.0 (!), and we  actively want to do migrations different way (https://github.com/ipfs/kubo/issues/10347#issuecomment-2388689056)

No need to keep poor outdated copy of a subset of [config.md#profiles](https://github.com/ipfs/kubo/blob/master/docs/config.md#profiles) Let's just remove this content.


2color (approved):
LGTM

Issue #1980, opened 2025-01-15
The documented JSON output for `/api/v0/dag/import` is incorrect


smoyer64:
The current documentation describes `/api/v0/dag/import` as returning a JSON body with the following structure:

```json
{
  "Root": {
    "Cid": {
      "/": "<cid-string>"
    },
    "PinErrorMsg": "<string>"
  },
  "Stats": {
    "BlockBytesCount": "<uint64>",
    "BlockCount": "<uint64>"
  }
}
```

What is actually returned is a JSONL body with the following structure:

```jsonl
"Root":{"Cid":{"/":"<cid-string>"},"PinErrorMsg":"string"}}
{"Stats":{"BlockCount":<uint64>,"BlockBytesCount":<uint64>}}
```

Note that the `Stats` "record" is only returned if the `stats=true` argument is provided.

lidel:
@smoyer64 thank you for filling this. 


As @2color  noted, these docs are generated by code mentioned in https://github.com/ipfs/ipfs-docs/pull/1981#issuecomment-2596004899

Your analysis from https://github.com/ipfs/ipfs-docs/pull/1981#issuecomment-2596444596 is spot-on, yes, it is intentional.

I think the only missing piece of the puzzle is streaming and why we have it and that CLI uses it by default. Some context below:

- `ipfs` CLI makes requests with streaming enabled for all commans by default (  `--stream-channels=true` is the implicit default and it makes HTTP client append `&stream-channels=true` to RPC requests).

- This makes RPC endpoint respond in streaming fashion (NDJSON-like), flushing after every object, allowing server to not have to buffer everything in memory before returning result, and client to consume and react to partial results before entire request is processed.

  <details>
    <summary>Click to expand `ipfs dag import` example</summary>
  
  You can see what happens on the wire bit better with some mitm like socat:
  
  ```console
  $ echo -n "hello" | ipfs block put --pin=true
  $ ipfs dag export bafkreibm6jg3ux5qumhcn2b3flc3tyu6dmlb4xa7u5bf44yegnrjhc4yeq > hello.car
  $ ipfs dag import --pin-roots=true --stats --api /dns/localhost/tcp/7423 hello.car
  ``` 
  
  And then socat proxy shows what happens on the wire bit better:
  
  ```console
  $ socat -v TCP-LISTEN:7423,fork TCP:localhost:5001
  [... sending CAR ... ]
  < 2025/01/16 22:48:40.000221008  length=566 from=0 to=565
[...]

PR #1981, opened 2025-01-15
docs(kubo-rps): describe JSONL returned by /api/vo/dag/import

smoyer64:
# Describe your changes
As described in #1980, the JSON body for the `/api/v0/dag/import` command is actually JSONL is the `stats=true` argument is included.

# Files changed

- docs/reference/kubo/rpc.md

#1980

- 
- 
- 

No

- 
- 

## Checklist before merging

### DIFF
--- a/docs/reference/kubo/rpc.md
+++ b/docs/reference/kubo/rpc.md
@@ -1102,22 +1102,11 @@ Argument `path` is of file type. This endpoint expects one or several files (dep
 
 ### Response
 
-On success, the call to this endpoint will return with 200 and the following body:
+On success, the call to this endpoint will return with 200 and the following JSONL body (if the `stats=true` argument is included):
 
 ```json
-{
-  "Root": {
-    "Cid": {
-      "/": "<cid-string>"
-    },
-    "PinErrorMsg": "<string>"
-  },
-  "Stats": {
-    "BlockBytesCount": "<uint64>",
-    "BlockCount": "<uint64>"
-  }
-}
-
+{"Root":{"Cid":{"/":"<cid-string>"},"PinErrorMsg":"string"}}
+{"Stats":{"BlockCount":<uint64>,"BlockBytesCount":<uint64>}}
 ```
 
 ### cURL Example


2color:
Thanks for the PR @smoyer64. I can confirm that two separate JSON objects are returned, in what seems to be like `ndjson`, though I'm not sure if that's actually deliberate. 

Either way, this can't be merged, because these docs are actually autogenerated: https://github.com/ipfs/ipfs-docs/blob/ff99b4d0c6ee2cd4acdf2204a83b0bf390014844/docs/reference/kubo/rpc.md?plain=1#L50-L53


@lidel I haven't been able to figure out exactly how the docs are generated, but it seems it's a combination of:
- https://github.com/ipfs/ipfs-docs/tree/main/tools/http-api-docs
- https://github.com/ipfs/go-ipfs-cmds
- https://github.com/ipfs/kubo/blob/f41b190e1b526ea2baa37f1ab991b7eba7b48790/core/commands/dag/dag.go#L171-L251

smoyer64:
> I AM SERIOUS, DO NOT EDIT ANYTHING BELOW ;-D

Well that was an epic fail ... I searched for the title line for that command and never scrolled up to see I was about to fail :)

I'll amend this PR (or replace it) once I figure out how the code generation works.  I believe this file was generated by the tool(s) at the first link you referenced ... and it appears to be keyed off this output type: https://github.com/ipfs/kubo/blob/f41b190e1b526ea2baa37f1ab991b7eba7b48790/core/commands/dag/dag.go#L205

That type is defined as https://github.com/ipfs/kubo/blob/f41b190e1b526ea2baa37f1ab991b7eba7b48790/core/commands/dag/dag.go#L63-L66

Which makes me think that

> though I'm not sure if that's actually deliberate.

might be very prescient.  I ran the equivalent command using the IPFS CLI and captured the response packet using WireShark.  The response is two chunks each with a complete JSON document:

image

I'll wait for @lidel to respond so I don't waste time chasing this.

P.S.  I asked my team members and they're vaguely remembering other places where the response is JSONL - so far, we haven't found those examples (yet).

smoyer64:
This code also makes me think this behavior is intentional.  And I therefore think that the problem is that the `CarImportOutput` type is NOT actually used to produce the response and that the problem is that the documentation generation tool can't represent the output of [this code](https://github.com/ipfs/kubo/blob/f41b190e1b526ea2baa37f1ab991b7eba7b48790/core/commands/dag/import.go#L151-L195), which will generate a JSONL (aka ndJSON) "record" for each CAR file's 'Root' as well as an optional record for the `Stats`

lidel:
@smoyer64 your analysis is correct, I've added more context in https://github.com/ipfs/ipfs-docs/issues/1980#issuecomment-2597040484.


 I'm closing this PR because the fix needs to happen elsewhere.
 Let's continue in #1980

PR #1996, opened 2025-02-18
fix: remove 32-bit binaries and ipfs-updater

lidel:
This PR

- [x] removes mentions of 32-bit binaries (https://github.com/ipfs/distributions/issues/1128)
- [x] removes mention of unmaintained `ipfs-updater` page
  - the tool is confusing users ([example](https://discuss.ipfs.tech/t/should-i-use-ipfs-update/19013?u=lidel)) and for a good reason (we don't maintain it, and it can break any time)


docs/install/command-line.md
```diff
@@ -342,6 +335,14 @@ When an IPFS command executes without parameters, the CLI client checks whether
 
 - If an `$IPFS_PATH` isn't in the default location, use the `--api <rpc-api-addr>` command-line argument. Alternatively, you can set the environment variable to `IPFS_PATH`. `IPFS_PATH` will point to a directory with the `$IPFS_PATH/api` file pointing at the Kubo RPC of the existing `ipfs` daemon instance.
 
+::: tip
+
+If you plan to expose safe subset of RPC API to the public internet with TLS encryption and HTTP authentication, check out the [TLS and HTTP Auth for Kubo with Caddy](../how-to/kubo-rpc-tls-auth.md) guide.
```
2color, line 340:
```suggestion
If you plan to expose the RPC API to the public internet with TLS encryption and HTTP authentication, check out the [TLS and HTTP Auth for Kubo with Caddy](../how-to/kubo-rpc-tls-auth.md) guide.
```

docs/install/command-line.md
```diff
@@ -342,6 +335,14 @@ When an IPFS command executes without parameters, the CLI client checks whether
...
+
+If you are looking for HTTP Fetch API designed for browsers and public internet, with [proper HTTP Cache Control](https://specs.ipfs.tech/http-gateways/path-gateway/#cache-control-response-header), see implementation-agnostic [HTTP Gateway](../reference/http/gateway.md) instead. It can be used in web apps thanks to [Verified Fetch](https://www.npmjs.com/package/@helia/verified-fetch)
```
2color, line 342: I think the verified fetch part is a bit misleading. 
```suggestion
If you are just interested in retrieval, see implementation-agnostic [HTTP Gateway](../reference/http/gateway.md) instead.
```

docs/.vuepress/redirects
```diff
@@ -43,14 +43,15 @@
...
+/how-to/ipfs-updater /install/command-line
 /install/command-line-quick-start/ /how-to/command-line-quick-start
 /install/js-ipfs/ https://github.com/ipfs/helia/wiki
 /introduction/ /concepts
 /introduction/faq/ /concepts/faq
 /introduction/how-ipfs-works/ /concepts/how-ipfs-works
 /introduction/overview/ /concepts/what-is-ipfs
 /introduction/usage/ /how-to/command-line-quick-start
-/install/ipfs-updater /how-to/ipfs-updater
+/install/ipfs-updater /how-to/ipfs-updater/
```
2color, line 54: Why not redirect straight to an existing page?
```suggestion
/install/ipfs-updater /install/command-line
```

PR #2001, opened 2025-02-25
ci: add workflow to deploy docs to storacha

2color:
## What

Add a workflow to build and deploy to Storacha using IPFS Deploy Action


lidel (commented):
:+1:

Issue #704, opened 2021-03-19
Clarify how to use the `x-directory` header in Curl requests.


autonome:
From a hackathon participant:

> I am trying to add a file in IPFS inside a specific directory via the HTTP API, can someone point me to an example in cURL?
curl -F "test.json=@/2021/03/18/weather.json" 127.0.0.1:5001/api/v0/add?wrap-with-directory=true
the above does not seem to create a "weather.json" file inside a directory structure "/2021/03/18" but instead it creates a file called "/2021/03/18/weather.json"

One answer I got when asking around:

> The api docs explain how the multipart body works. You cannot let curl build the multipart as normal, since you need to send a x-directory part apart from the uploaded file part i think, or at least set some part header manually with the path. I never got to figure out what curl foo magic is needed and whether this is possible.

If we don't know how to use it, nobody else will either 😅

autonome:
Updated answer for future use in examples:

> curl -F 'file=@ipfsgif.gif;encoder=base64;filename=a/b/c.gif' 'http://localhost:5001/api/v0/add?progress=true'
{"Name":"a/b/c.gif","Bytes":33534}
{"Name":"a/b/c.gif","Hash":"QmVCTMQrvAj1Czk4beNxD5LKP4sHYFwTwPDvc5UDTaGzPB","Size":"33548"}
{"Name":"a/b","Hash":"Qmb8UKKhhqRUpgyhBK1HfdzNZKydRo5EwDj334BuUSLjPy","Size":"33601"}
{"Name":"a","Hash":"Qma4t85aY6LoGYbasRidumoMZizkhRbVHkk8EV2u7DLGYg","Size":"33650"}

And docs link for the API:

https://docs.ipfs.io/reference/http/api/#request-body

Could be an easy fix to close this gap is to just put the full curl example in that part of the docs?

autonome:
So, there are two things at play here, `x-directory` and `filename`:

> You can also put x-directory parts, they don't hurt, and are needed for example to create an empty dir I think, but otherwise things are inferred from the filename header.

johnnymatthews:
Just to make sure I'm understanding correctly: 

1. The current example just creates a file called `/2021/03/18/weather.json`:

    ```
    curl -F "test.json=@/2021/03/18/weather.json" 127.0.0.1:5001/api/v0/add?wrap-with-directory=true
    ```

2. The actual command should be:

    ```
[...]

PR #2012, opened 2025-03-26
Add folder to IPFS - helpful details

ProximaNova:
# Describe your changes
Adds an example on how to add a directory (not just a single file) by using curl and `/api/v0/add`. API docs are pretty clear as I've read about all of the page. The one unclear part is about adding folders.

# Files changed
- ipfs-docs/blob/main/tools/http-api-docs/markdown.go

- None. No GitHub issue/PR on this found. Maybe similar to adding directories: https://github.com/ipfs/ipfs-docs/issues/704 . About that, as far as I know, the only way to add a folder by using /api/v0/add is to specify one or multiple files plus the options in this PR. You cannot have a path to a folder and expect /api/v0/add to work. Or can you? (Something <b>like</b> `curl -XPOST -F "file=@folder/;filename=folder" 127.0.0.1:5001/api/v0/add`?) The documentation is unclear and this PR helps to fix that.

- No

## Checklist before merging

### DIFF
--- a/tools/http-api-docs/markdown.go
+++ b/tools/http-api-docs/markdown.go
@@ -350,7 +350,7 @@ Argument `+"`%s`"+` is of file type. This endpoint expects one or several files 
 		if bodyArg.Endpoint == "/api/v0/add" {
 			fmt.Fprintln(buf, `
 
-The `+"`add`"+` command not only allows adding files, but also uploading directories and complex hierarchies. (For example, in cURL, you can upload multiple files and have them show up in a folder by running `+"`curl -sLk -XPOST -F \"file=@file1.txt;filename=path1/file1.txt\" -F \"file=@file2.txt;filename=path2/file2.txt\" \"http://127.0.0.1:5001/api/v0/add?recursive=true&wrap-with-directory=true\"`"+`.)
+The `+"`add`"+` command not only allows adding files, but also uploading directories and complex hierarchies. (For example, in cURL, you can upload multiple files and have them show up in a folder by running `+"`curl -sLk -XPOST -F \"file=@file1.txt;filename=path1/file1.txt\" -F \"file=@file2.txt;filename=path2/file2.txt\" \"http://127.0.0.1:5001/api/v0/add?recursive=true&wrap-with-directory=true\"`"+`; -F is specified multiple times along with recursive and wrap-with-directory.)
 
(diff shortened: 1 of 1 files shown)


ProximaNova:
The "Files changed" tab shows a basic way of adding a folder via the API. I haven't come up with a great automated way to add any folder. Not really important for this PR but still worth noting - add current working directory to IPFS via the HTTP API:
```
# Works, but fails if whitespace is in any filename, no eval is used:
filesf() { basedir="$(pwd)"; basedirlen=$(echo "$basedir" | wc --bytes); find "$basedir" -type f | basedirlen="$basedirlen" xargs -d "\n" sh -c 'for args do nobasedir=$(echo "$args" | sed -E "s/^.{$basedirlen}//g"); echo " -F "file=@"\"$args\";filename=\"$nobasedir\"" | tr -d \\n; done' _; }
curl -sLk -XPOST -H "Content-Type: multipart/form-data" $(filesf) "127.0.0.1:5001/api/v0/add?cid-version=1&chunker=size-1048576&recursive=true&wrap-with-directory=true&pin=false"

# Works on any filename, but uses eval
eval $(echo "arr=($(basedir="$(pwd)"; basedirlen=$(echo "$basedir" | wc --bytes); find "$basedir" -type f | basedirlen="$basedirlen" xargs -d "\n" sh -c 'for args do nobasedir=$(echo "$args" | sed -E "s/^.{$basedirlen}//g"); echo "\"file=@$args;filename=$nobasedir\" " | tr -d \\n; done' _ | sed "s/ $//g"))")
toexe="$(i=0; text='curl -sLk -XPOST -H "Content-Type: multipart/form-data"'; while [[ i -lt ${#arr[@]} ]]; do text="$text -F \"${arr[$i]}\""; echo $text; ((i++)); done | tail -n1 | sed "s/$/ \"127.0.0.1:5001\/api\/v0\/add?cid-version=1\&chunker=size-1048576\&recursive=true\&wrap-with-directory=true\&pin=false\"/g")"
eval $(echo "$toexe")
```

2color:
This file shouldn't be edited directly.

https://github.com/ipfs/ipfs-docs/blob/c134198b7beff0dc365d6bcca4fc72d1da43c1de/tools/http-api-docs/markdown.go#L26-L34

lidel:
Let's continue in https://github.com/ipfs/ipfs-docs/pull/2014

PR #2116, opened 2025-08-24
ci: modernize workflows to support fork pr previews

lidel:
splits workflow to enable ipfs deployments for all pull requests including those from forks

- build.yml: builds site and uploads artifacts
- deploy.yml: handles ipfs deployment via workflow_run
- uses ipfs-deploy-action@b491fdc with workflow_run support from https://github.com/ipshipyard/ipfs-deploy-action/pull/37

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,28 +1,38 @@
-name: Build and Deploy to IPFS
+# Build workflow - runs for both PRs and main branch pushes
+# This workflow builds the website without access to secrets
+# For PRs: Runs on untrusted fork code safely (using pull_request event, not pull_request_target)
+# For main: Builds and uploads artifacts for deployment
+# Artifacts are passed to the deploy workflow which has access to secrets
+
+name: Build
 
 permissions:
   contents: read
-  pull-requests: write
-  statuses: write
+
 on:
   push:
     branches:
       - main
-  pull_request_target:
+  pull_request:
     branches:
       - main
 
+env:
+  BUILD_PATH: 'docs/.vuepress/dist'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
-  build-and-deploy:
+  build:
     runs-on: ubuntu-latest
-    outputs: # This exposes the CID output of the action to the rest of the workflow
-      cid: ${{ steps.deploy.outputs.cid }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || github.sha }}
+        # No ref parameter needed - uses correct ref automatically:
+        # - For PRs: merge commit or PR head
+        # - For pushes: the pushed commit
 
       - name: Setup Node.js
         uses: actions/setup-node@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -1,28 +1,38 @@
...
     runs-on: ubuntu-latest
-    outputs: # This exposes the CID output of the action to the rest of the workflow
-      cid: ${{ steps.deploy.outputs.cid }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || github.sha }}
+        # No ref parameter needed - uses correct ref automatically:
+        # - For PRs: merge commit or PR head
```
2color: How is it decided if it's the merge commit or PR head?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,58 @@
...
+  pull-requests: write
+  statuses: write
+
+on:
+  workflow_run:
+    workflows: ["Build"]
+    types: [completed]
+
+env:
+  BUILD_PATH: 'docs-build'
```
2color, line 21: Where is the `docs-build` constant coming from?

.github/workflows/build.yml
2color: [According to the docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request), it defaults to the merge commit, which _may seem confusing_, because it doesn't reflect the actual HEAD state of the PR branch. Instead, it contains PR changes PLUS any new commits that landed on main since the PR was created.

.github/workflows/deploy.yml
lidel: just arbitrary name -- in `deploy.yml` we download  and unpack `build.yml` result to  `BUILD_PATH` and then `ipfs-deploy-action` reads from it

.github/workflows/build.yml
lidel: good catch, i made a mental shortcut, clarified in 3872fec3c45dd32d9ef52c7bfe0258a757d01176

.github/workflows/build.yml
```diff
@@ -1,28 +1,38 @@
...
+  build:
     runs-on: ubuntu-latest
-    outputs: # This exposes the CID output of the action to the rest of the workflow
-      cid: ${{ steps.deploy.outputs.cid }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.sha || github.sha }}
+        # No ref parameter needed - uses correct ref automatically:
```
2color, line 33: This does change the behaviour to checkout the merge commit. I guess there's a reason for this being the default. So let's give it a go and change if needed in the future.

2color (approved):

.github/workflows/build.yml
2color, line 33: The problem with this, which I just realised, is that this will cause ipfs-deploy-action to set the commit status for merge SHA commits that aren't visible in the pull request. 

Both Cloudflare Pages and Vercel use the PR head commit for the following reasons:

- The preview deployment reflects exactly what's in your feature branch
- It doesn't include any potential merge conflicts or changes that would result from merging with the target branch
- Each new commit pushed to the PR branch triggers a new preview deployment from that latest commit


If we want to ensure that PRs are up to date and avoid merge conflicts, we can require branches to be up to date before merging in the branch protection rules in GitHub.

2color: I'm currently testing this in this PR: https://github.com/ipfs/ipfs-docs/pull/2119

2color: For some reason, with #2119 the commit SHA that was used in the PR comment is `pull_request.head.sha` while the code that was checked out was technically the merge commit SHA (which should be different).

I'm currently investigating why.

2color: I can confirm now that there's a discrepancy between the commit SHA used for the build and the commit SHA used to update the comment. 

You can see it here: https://github.com/ipfs/ipfs-docs/actions/runs/17208024892/job/48812841343#step:2:87


It's checking out and building e8daa379e083016a0c35bce97da1ad519c284287, but setting the commit status for a3bd938afef2529cc8a9895e203159139e37c5d4.

PR #2119, opened 2025-08-25
Remove outdated content provider list and update peering documentation

2color:
## Summary

- Remove outdated content provider list section
- Update peering documentation with improved clarity and accurate links

## Motivation

The content provider list contained many peers that are no longer online. Rather than maintaining a list that's bound to get outdated and mislead users, this PR removes the provider list entirely and focuses the documentation on the core peering concepts and configuration.

More broadly, [reprovide sweep](https://github.com/libp2p/go-libp2p-kad-dht/issues/824) should bring significant improvements to the DHT, which should eliminate many of the content routing problems experienced by the community.


2color:
> There is far too much tribal knowledge present in the IPFS ecosystem. Removing this list is a step in the wrong direction.

Can you please explain the reasoning?

PR #2126, opened 2025-08-28
docs: make explicit that you can use `ipfs cid format` to convert both ways

dannyob:
It took me a moment to realise that this was possible, so wanted to save that moment for future generations of IPFS newbies.

# Describe your changes
Just an extra line in the docs explicitly showing how to use ipfs cid format -v 0.

# Files changed
- docs/concepts/content-addressing.md

nope!

## Checklist before merging

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - id: checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
     - id: link-check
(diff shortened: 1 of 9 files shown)


lidel (approved):
Thank you @dannyob 

Caveat: conversion from CIDv1 to CIDv0 is only possible when the CIDv1 uses the dag-pb codec (0x70), as CIDv0 only supports dag-pb. CIDs using other codecs like raw (0x55), dag-cbor (0x71), or dag-json (0x0129) cannot be converted to CIDv0. Added this clarification.

PR #2216, opened 2025-12-09
docs: add quick download links to landing pages

### LINKED ISSUE #2129: Link para download
Na pagina não tem o link para download

lidel:
users landing on docs homepage or install page often want to download IPFS quickly. previously they had to navigate through multiple pages to find actual download links.

added tip boxes at the top of both pages with direct links to install sections (via anchor fragments), so users who know what they want can get there immediately while preserving the existing page structure for those who need to explore options.

Closes #2129

### DIFF
--- a/mlc_pull_req_config.json
+++ b/mlc_pull_req_config.json
@@ -10,6 +10,9 @@
     "ignorePatterns": [
         {
             "pattern": "^[^/]+$"
+        },
+        {
+            "pattern": "^https://www.npmjs.com/"
         }
     ],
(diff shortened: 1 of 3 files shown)


mlc_pull_req_config.json
```diff
@@ -10,6 +10,9 @@
     "ignorePatterns": [
         {
             "pattern": "^[^/]+$"
+        },
+        {
+            "pattern": "^https://www.npmjs.com/"
```
lidel, line 15: ℹ️  this is important, because without this every new PR will break. merging

PR #2243, opened 2026-01-23
Add scientific data and geospatial publishing guides & case study

2color:
## What

Adds new documentation focused on scientific/geospatial data publishing with IPFS (Zarr + tooling), plus an ORCESTRA case study and related VuePress navigation updates, with a small quickstart retrieval enhancement.

Changes:

Added “Scientific data and IPFS landscape guide” and “Publish geospatial Zarr data with IPFS” how-to pages.
Added ORCESTRA case study and updated VuePress sidebar/navigation to include the new section + case study.
Extended retrieval quickstart with a Python/ipfsspec verified retrieval example and updated spellcheck ignore list.


docs/how-to/publish-geospatial-data.md
```diff
@@ -0,0 +1,216 @@
...
+- **Participatory sharing**: Anyone can help distribute data sets they've downloaded
+- **Open access**: No vendor lock-in or centralized infrastructure required
+
+This combination has proven effective in real-world campaigns like [Orcestra](https://orcestra-campaign.org/orcestra.html), where scientists collaborated with limited internet connectivity in the field while sharing data globally.
+
+## Prerequisites
+
+Before starting, ensure you have:
+
+- A Zarr data set for
```
mishmosh: Fragment?

docs/how-to/publish-geospatial-data.md
```diff
@@ -0,0 +1,216 @@
...
+- [Kubo](/install/command-line/) or [IPFS Desktop](/install/ipfs-desktop/) installed on a machine with a public IP
```
mishmosh: > machine with a public IP

can we link to a resource to help people check if they have a public IP, or configure one if not?

docs/how-to/publish-geospatial-data.md
```diff
@@ -0,0 +1,216 @@
...
+         --pin-name="halo-measurements-2026-01-23" \
+         --quieter \
+         ./my-dataset.zarr
+```
+
+This command:
+
+1. **Merkleizes** the folder: converts files and directories into content-addressed blocks with UnixFS
+2. **Pins** the data locally: prevents garbage collection from removing it
+3. **Queues providing**: announces to the IPFS network that your node has this data
```
mishmosh: "Queues providing" introduces multiple new terms at once. Would any of these alternates work?
- Broadcasts availability
- Starts providing
- Starts advertising
- Advertises the content

docs/how-to/publish-geospatial-data.md
```diff
@@ -0,0 +1,216 @@
...
+
+```bash
+ipfs stats provide
+```
+
+For detailed diagnostics, see the [provide system documentation](https://github.com/ipfs/kubo/blob/master/docs/provide-stats.md).
+
+## Step 5: Content Discovery
+
+Users need a way to discover your datasets. Choose an approach based on your needs:
```
mishmosh:
```suggestion
Now that your data is available on the public network, the next step is making it discoverable to others. Choose a sharing approach based on your needs:
```

docs/how-to/publish-geospatial-data.md
```diff
@@ -0,0 +1,216 @@
...
+
+For one-off sharing, provide the CID directly:
+
+```
+ipfs://bafybeif52irmuurpb27cujwpqhtbg5w6maw4d7zppg2lqgpew25gs5eczm
+```
+
+### Option B: Use IPNS for Updatable References
+
+IPNS provides a stable identifier that you can update when datasets change:
```
mishmosh:
```suggestion
If you want to share a stable identifier but be able to update the underlying dataset, create an [IPNS](https://docs.ipfs.tech/concepts/ipns/) identifier and share that instead. This is useful for datasets that get updated regularly —users can bookmark your IPNS name and always retrieve the latest version.
```

(not sure if that's the correct way to link internally)

docs/how-to/publish-geospatial-data.md
```diff
@@ -0,0 +1,216 @@
...
+
+```bash
+# Publish your dataset under your node's IPNS key
+ipfs name publish /ipfs/<your-dataset-cid>
+
+# Update to a new version later
+ipfs name publish /ipfs/<new-dataset-cid>
+```
+
+Users can subscribe to your IPNS name to always get the latest version.
```
mishmosh: Is this an explicit, persistent "subscribe" function? Or do we recommend using it like an IPFS CID (in which case maybe we should describe this without the word "subscribe")

docs/how-to/publish-geospatial-data.md
```diff
@@ -0,0 +1,216 @@
...
+```python
+import xarray as xr
+
+ds = xr.open_dataset(filename)
+# Example: targeting ~1 MB chunks with float32 data
+ds.to_zarr('output.zarr', encoding={
+    'var_name': {'chunks': (1, 512, 512)}
+})
+
+# Total size: 1 × 512 × 512 × 4 bytes (float32) = ~1 MB per chunk
```
vmx:
```suggestion
# Total size: 1 × 512 × 512 × 4 bytes (float32) = 1048576 bytes = 1 MiB per chunk
```
I would add the "1048576 bytes" explicitly, so that the `chunker-size` below makes sense (you could search the page for that number and find where it comes from).

docs/how-to/publish-geospatial-data.md
```diff
@@ -0,0 +1,216 @@
...
+This command:
+
+1. **Merkleizes** the folder: converts files and directories into content-addressed blocks with UnixFS
+2. **Pins** the data locally: prevents garbage collection from removing it
+3. **Queues providing**: announces to the IPFS network that your node has this data
+4. **Outputs the root CID**: the identifier for your entire dataset
+
+The `--quieter` flag outputs only the root CID, which identifies the complete dataset.
+
+> **Note:**
```
vmx: Empty note, I guess it's some left-over.

docs/how-to/scientific-data/publish-geospatial-zarr-data.md
```diff
@@ -0,0 +1,216 @@
...
+
+**Python with ipfsspec**:
+
+```python
+import xarray as xr
+
+ds = xr.open_dataset(
+    "ipfs://bafybeif52irmuurpb27cujwpqhtbg5w6maw4d7zppg2lqgpew25gs5eczm",
+    engine="zarr"
+)
```
vmx, line 277: A link to ipfsspec would be good.
For those (like me) that aren't that familiar with fsspec, it might be good to add a comment in the code like ```# after the installation of ipfsspec, `ipfs://` urls are automatically recognized```

docs/how-to/publish-geospatial-data.md
2color: Yeah this is a gap that is best addressed by a new command in Kubo rather than a guide (because to test this you need some external node checking that you are in fact reachable). 


I opened https://github.com/ipfs/kubo/issues/11171 to address this.

2color: And opened https://github.com/ipfs/kubo/pull/11184 to fix it

2color: I opened https://github.com/ipfs/ipfs-docs/issues/2251 to track the future guide for this. 

Also going to add a note about port mapping as an alternative solution. 

Resolving this issue for now.

docs/how-to/publish-geospatial-data.md
2color: Good point. We don't have a subscribe function. I'll update to reflect that.

docs/how-to/scientific-data/publish-geospatial-zarr-data.md
2color: Addressed

docs/how-to/scientific-data/publish-geospatial-zarr-data.md
```diff
@@ -0,0 +1,233 @@
...
+
+Link a DNS name to your content by adding a TXT record:
+
+```
+_dnslink.data.example.org  TXT  "dnslink=/ipfs/<cid>"
+```
+
+Users can then access your data by using the `ipns://` prefix.
+
+TODO: add example.
```
2color: Note to self: add example

docs/how-to/scientific-data/publish-geospatial-zarr-data.md
```diff
@@ -0,0 +1,233 @@
...
+## Step 2: Add Your Data Set to IPFS
+
+Add your Zarr folder to IPFS using the `ipfs add` command:
+
+```bash
+ipfs add --recursive \
+         --hidden \
+         --raw-leaves \
+         --chunker=size-1048576 \
+         --cid-version=1 \
```
lidel: Once Kubo 0.40 ships these could be removed and replaced by one-time `ipfs config profile apply unixfs-v1-2025` or setting `Import.*` values one-by-one

vmx (approved):

docs/how-to/scientific-data/landscape-guide.md
```diff
@@ -0,0 +1,279 @@
...
+
+# Scientific Data and IPFS Landscape Guide
+
+<!-- ## Landscape and Ecosystem Tooling -->
+
+Scientific data and IPFS are naturally aligned: research teams need to share large datasets across institutions, verify data integrity, and ensure resilient access. From sensor networks to global climate modeling efforts, scientific communities are using IPFS content addressing and peer-to-peer distribution to solve problems traditional infrastructure can't.
+
+In this guide, you'll find an overview of the problem space, available tools, and architectural patterns for publishing and working with scientific data using IPFS.
+
+## A Landscape in Flux
```
2color: From Slack:

> Condense “A Landscape in Flux” - Remove or significantly shorten so we get to solutions/new knowledge faster; we can highlight the main points but you can assume readers already work in geospatial and are familiar with the challenges

docs/case-studies/orcestra.md
```diff
@@ -0,0 +1,126 @@
...
+The campaign brings together **over twenty scientific institutions** spanning Europe, North America, and Africa. Eight sub-campaigns (three airborne, one land-based, and four at sea) coordinate aircraft, ships, ground stations, and satellites to collect atmospheric measurements across the tropical Atlantic.
+
+ORCESTRA represents the kind of large-scale scientific collaboration where data infrastructure can make or break a mission: dozens of research groups generating terabytes of observational data that must be shared, verified, and preserved across institutional boundaries.
+
+### ORCESTRA by the numbers
+
+<NumberBlock :items="[
+  {value: '20+', text:'Research institutions'},
+  {value: '8', text: 'Sub-campaigns'},
+  {value: '50+', text: 'Participating organizations'},
```
lkluft:
```suggestion
  {value: '50+', text: 'Scientists on-site'},
```

docs/case-studies/orcestra.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Overview
+
+In this case study, you'll learn how [The ORCESTRA Campaign](https://orcestra-campaign.org/), a coordinated international atmospheric science campaign, uses IPFS to ensure that scientific datasets are verifiable, openly accessible, and collaboratively distributed across institutions worldwide.
+
+## What is ORCESTRA
+
+[ORCESTRA](https://orcestra-campaign.org/) (Organized Convection and EarthCARE Studies over the Tropical Atlantic) is an international field campaign that launched in early 2024 to study tropical mesoscale convective systems: the storm systems that play a significant role in the Earth's weather and climate dynamics.
+
+The campaign brings together **over twenty scientific institutions** spanning Europe, North America, and Africa. Eight sub-campaigns (three airborne, one land-based, and four at sea) coordinate aircraft, ships, ground stations, and satellites to collect atmospheric measurements across the tropical Atlantic.
```
lkluft: While we did perform overflights with several satellites, one might mention EarthCARE explicitly (as it also comes up in the name).

```suggestion
The campaign brings together **over twenty scientific institutions** spanning Europe, North America, and Africa. Eight sub-campaigns (three airborne, one land-based, and four at sea) coordinate aircraft, ships,  and ground stations to collect atmospheric measurements across the tropical Atlantic and validate observations made by the [EarthCARE](https://earth.esa.int/eogateway/missions/earthcare) satellite, which was launched shortly before the start of the campaign.
```

docs/case-studies/orcestra.md
```diff
@@ -0,0 +1,126 @@
...
+
+A Raspberry Pi with a 1 TB SD card became that solution. Running a [Kubo](https://github.com/ipfs/kubo) IPFS node, the device enabled local data sharing from laptop to laptop, provided temporary storage on the Pi node, and facilitated eventual transfer to a data center in Hamburg. No central server required; just content-addressed, peer-to-peer data sharing that worked.
+
+This improvised setup revealed something important: IPFS wasn't just a workaround, it was a better fit for how field science actually works. As the campaign evolved and datasets grew, the team expanded IPFS from an emergency fix to the core of their data publishing infrastructure.
+
+## How ORCESTRA works
+
+ORCESTRA's eight sub-campaigns span sea, air, and land, collecting atmospheric measurements such as temperature, humidity, wind, radiation, aerosols, and cloud properties. This observational data is structured as multidimensional arrays and stored primarily in the [Zarr](https://zarr.dev/) format, a cloud-native format optimized for chunked, distributed access to large scientific datasets.
```
lkluft:
```suggestion
ORCESTRA's eight sub-campaigns cover sea, air and land. They collect atmospheric measurements such as temperature, humidity, wind, radiation, aerosols and cloud properties, as well as oceanic measurements such as sea-surface temperature, salinity and ocean currents. This observational data is structured as multidimensional arrays and stored primarily in the [Zarr](https://zarr.dev/) format, a cloud-native format optimized for chunked, distributed access to large scientific datasets.
```

lkluft (approved):
Thanks for adding this really nice article about ORCESTRA!

I made a couple of minor suggestions, but I do like the overall story very  much! 👍

docs/how-to/scientific-data/publish-geospatial-zarr-data.md
```diff
@@ -0,0 +1,293 @@
...
+## Step 1: Prepare your Zarr data set
+
+When preparing your Zarr data set for IPFS, aim for approximately 1 MiB chunks to align with IPFS's 1 MiB maximum block size. While this is not a strict requirement, using larger Zarr chunks will cause IPFS to split them into multiple blocks, potentially increasing retrieval latency.
+
+To calculate chunk dimensions for a target byte size, work backwards from your datatype:
+
+```python
+import xarray as xr
+
+ds = xr.open_dataset(filename)
```
vmx, line 82: In the MFS section you introduce "HALO flight datasets". I think it would be good if you could introduce them right here at the beginning and then make all steps working with that dataset, so one can follow along locally.

When I started to look into Zarr, I had a hard time figuring out what it exactly looked like on disk. Most resources I found where online notebooks. So such a start to finish walk through with actual data would've been useful. E.g. it would've been clearer to me that a "somedata.zarr" is actually a directory and not a single file.

Also mentioning that this is about Zarr v3 would make sense as Zarr v2 still seems to be quite common.

docs/how-to/scientific-data/publish-geospatial-zarr-data.md
```diff
@@ -0,0 +1,293 @@
...
+# added bafybeiheegb3h5s4v4igmxqivp4mq5gvp4zyv6lc3nmjjifgybqovzh4ji  halo-measurements-2026-01-23
+
+# Link the two remote datasets into the same MFS directory —
+# the remote data stays on the network, no local download required
+ipfs files cp /ipfs/bafybeiaxkxdnj7wfivrkjw7qhglcvbulryvpxp5cqr7djjdnywefr4cmq4 /datasets/halo/2026-01-21
+ipfs files cp /ipfs/bafybeibpfqhzuicdpfzzfujkzaqnxz5e7c5e44erjpxjrguqxs7y6bcnve /datasets/halo/2026-01-22
+
+# Each mutation produces a new root CID — a lightweight versioned snapshot
+ipfs files stat --hash /datasets/halo
+# bafybeihqixf5ew7mfr74bzb74qiw2mgtnytabnpzjnf5xeejzq4p2ocygu
```
vmx, line 191: Maybe add a final `ls` to show that all directories are there.

docs/quickstart/retrieve.md
```diff
@@ -92,6 +93,37 @@ You may notice that there's a path following the CID, e.g. `bafybeicn7i3soqdgr7d
 :::
 
 
+## Fetching the CID with Python and ipfsspec
+
+[ipfsspec](https://github.com/fsspec/ipfsspec) is a read-only [fsspec](https://filesystem-spec.readthedocs.io/) implementation for IPFS. It performs **verified** retrieval by fetching [CAR files](../concepts/glossary.md#car) containing Merkle proofs, so you don't have to trust the gateway. It works without a local IPFS node.
```
vmx: I'd like to have it mention that it fetches from a gateway. It should be clear implicitly, but I think mentioning it explicitly is better.

docs/how-to/scientific-data/publish-geospatial-zarr-data.md
```diff
@@ -0,0 +1,293 @@
...
+# Two datasets already exist on the network from other contributors.
+# Their root CIDs are known (e.g. from a manifest or data catalogue):
+#   2026-01-21 flight → bafybeiaxkxdnj7wfivrkjw7qhglcvbulryvpxp5cqr7djjdnywefr4cmq4
+#   2026-01-22 flight → bafybeibpfqhzuicdpfzzfujkzaqnxz5e7c5e44erjpxjrguqxs7y6bcnve
+
+# Add the new local flight dataset, placing it directly into MFS
+ipfs add --recursive ./halo-measurements-2026-01-23 --to-files=/datasets/halo/2026-01-23
+# added bafybeiheegb3h5s4v4igmxqivp4mq5gvp4zyv6lc3nmjjifgybqovzh4ji  halo-measurements-2026-01-23
+
+# Link the two remote datasets into the same MFS directory —
```
2color, line 184: As it turns out, Kubo needs to fetch the root block in order to do these `cp` operations, so it's not a purely offline operation.

I'd need to overhaul this example to use real CIDs that are retrievable for it to work and update

docs/how-to/scientific-data/publish-geospatial-zarr-data.md
mishmosh: @2color kubo 0.40 shipped yesterday so this can be updated now!

docs/how-to/scientific-data/publish-geospatial-zarr-data.md
2color: I'll address this in a followup. I opened an issue: https://github.com/ipfs/ipfs-docs/issues/2260

PR #2262, opened 2026-03-06
Add guide about content addressing data sets

2color:
# Describe your changes

This is a new comparison guide of UnixFS, iroh collections, and DASL/MASL for content addressing directories of files, covering overhead, determinism, subsetting, and ecosystem support.

After spending a lot of time comparing the three for a different use case, I thought it would be useful to share these insights, and embrace the plurality of the IPFS ecosystem. 

## Checklist before merging
- [x] Passing all required checks (The beta **Check Markdown links for modified files** check is not required)

### DIFF
--- a/docs/how-to/README.md
+++ b/docs/how-to/README.md
@@ -5,20 +5,13 @@ description: Hands-on guides to using and developing with IPFS to build decentra
 
 # IPFS Guides and Tutorials
 
-::: callout
-Check out the new [Ecosystem guides](#ecosystem-guides) section to learn more about the amazing tools, software and implementations created by IPFS ecosystem partners.
-:::
-
-No matter what you're looking to do with IPFS, you can find how-tos and tutorials here. These items are a work in progress, so please check back periodically to check what we've added!
-
 See the site navigation menu for all our how-tos, organized by topic area, including favorites like these:
 
 - **Customize your install** by [configuring a node](configure-node.md), modifying the [bootstrap list](modify-bootstrap-list.md), and more
-- **Learn how to manage files** in IPFS with tutorials on concepts like [pinning](pin-files.md), how to [work with blocks](work-with-blocks.md), learning how to [troubleshoot file transfers](https://github.com/ipfs/kubo/blob/master/docs/file-transfer.md), and understanding [working with large datasets](https://github.com/ipfs/archives/tree/master/tutorials/replicating-large-datasets)
-- **See how to work with peers** using methods like [customizing libp2p bundles](https://github.com/ipfs-examples/js-ipfs-examples/tree/master/examples/custom-libp2p) and using circuit relay
+- **Learn how to manage files** in IPFS with tutorials on concepts like [pinning](pin-files.md), how to [work with blocks](work-with-blocks.md), learning how to [content address data sets](content-addressing-data-sets.md).
+- **Publish scientific data** by exploring the [scientific data and IPFS landscape guide](scientific-data/landscape-guide.md) or learning how to [publish geospatial Zarr data with IPFS](scientific-data/publish-geospatial-zarr-data.md)
 - **Understand website hosting** by starting with how to [host a simple single-page site](websites-on-ipfs/single-page-website.md)
-- **Learn how to build apps** on IPFS, starting with [exploring the IPFS API](https://github.com/ipfs/camp/tree/master/CORE_AND_ELECTIVE_COURSES/CORE_COURSE_C) and [making a basic libp2p app](https://github.com/ipfs/camp/tree/master/CORE_AND_ELECTIVE_COURSES/CORE_COURSE_B)
-- **Understand how IPFS works in the browser** by learning how to [address IPFS on the Web](address-ipfs-on-web.md) and [how IPFS can be used in your favorite browser tools and frameworks](browser-tools-frameworks.md)
+- **Understand how to use IPFS in the browser** by learning how to [address IPFS on the Web](address-ipfs-on-web.md) and [IPFS in web applications](ipfs-in-web-apps.md)
 
 
 ## Don't see what you're looking for?


docs/how-to/content-addressed-folders.md
```diff
@@ -0,0 +1,188 @@
...
+
+- [UnixFS](https://specs.ipfs.tech/unixfs/)
+- [iroh collections](https://docs.iroh.computer/protocols/blobs#collections)
+- [DASL](https://dasl.ing) along with its metadata system [MASL](https://dasl.ing/masl.html)
+
+The goal is to have a single content hash that represents a directory of files, such that verifying that hash verifies the entire contents.
+
+This matters for build outputs, software distributions, large datasets, website archives — any case where you need to verify that a collection of files hasn't changed. A naive approach like hashing a tarball is fragile: tar archives encode metadata (timestamps, permissions, ordering) that vary between machines, producing different hashes for identical file contents. Content addressing solves this, but the choice of format has real consequences — particularly for overhead, determinism, language support and existing tooling, and whether you can fetch subsets without downloading the whole thing. These differences compound as dataset size grows: what's negligible at megabyte scale — a few extra bytes of framing, an extra round of parsing per block — becomes a meaningful cost at terabyte scale across millions of files.
```
2color: Add note here about this creating a tar isn't practical for large datasets where you cannot afford to store two copies.

docs/how-to/content-addressed-folders.md
```diff
@@ -0,0 +1,188 @@
...
+Serialized with [postcard](https://docs.rs/postcard):
+
+```
+┌──────────────────────────────┐
+│ header: "CollectionV0."      │  13 bytes, magic/version tag
+├──────────────────────────────┤
+│ names: Vec<String>           │  varint-prefixed length, then
+│   "assets/style.css"        │  each string is varint-length
+│   "js/app.js"               │  prefixed + raw UTF-8 bytes
+│   "index.html"              │
```
vmx: The right edge is one space off. Same on the diagram below.

docs/how-to/content-addressed-folders.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Characteristics
+
+- **No metadata pollution.** Unlike tar/zip, there are no timestamps, permissions, or ownership fields. Two directories with identical file names and contents always produce the same hash, regardless of when or where they were produced.
+- **Flat representation of trees.** Directory structure lives in the name strings as relative paths, not as separate directory entries with their own metadata. One entry per file, no ambiguity about empty directories or nested paths.
+- **Positional, tag-free encoding.** Postcard serializes fields in declaration order with no field numbers or type tags. The `"CollectionV0."` magic header handles versioning.
+- **Compact.** The overhead per file is a varint-prefixed filename in the metadata blob and a 32-byte hash in the root blob.
+- **O(1) file lookup.** The root blob is a flat array of fixed-size 32-byte hashes, so finding the Nth file is a constant-time offset (`N * 32` bytes) with no parsing required.
```
vmx: When you already know at which offset the hash is. Practically you'd fetch the `HashSeq`, then the `CollectMeta` and scan though the strings to get the offset. It's still O(1), but it might be worth noting.

docs/how-to/content-addressed-folders.md
```diff
@@ -0,0 +1,188 @@
...
+- **Streaming verification.** The root blob is a hash sequence, so a verifier can check individual files incrementally as they arrive.
+- **Ready-made distribution.** Collections can be distributed in a peer-to-peer fashion with iroh-blobs.
+- **BLAKE3.** Fast (parallelizable, SIMD-accelerated), 256-bit digests, and adopted by the [BDASL](https://dasl.ing/bdasl.html) spec.
+- **File-level subsetting only.** Individual files can be fetched and verified by their BLAKE3 hash, but there is no way to address a subdirectory as a unit. Fetching a subset means filtering the path list and requesting files one by one.
```
vmx: This is kind of the same point as the "Flat representation of trees" above. Maybe those points can be combined.

I haven't checked, but couldn't a HasSeq also link to another HashSeq? If yes then you could build directory structures, where you can directly link to a sub-directory.

docs/how-to/content-addressed-folders.md
```diff
@@ -0,0 +1,188 @@
...
+- **Rust only (for now).** The reference implementation is in Rust. The format is simple enough to implement in other languages — it's just postcard-encoded strings and a flat array of BLAKE3 hashes — but no other implementations exist yet.
+
+## Comparison
+
+| Criteria             | iroh collections                                                        | UnixFS                                   | MASL/DRISL                                                                            |
+| -------------------- | ----------------------------------------------------------------------- | ---------------------------------------- | ------------------------------------------------------------------------------------- |
+| Encoding             | Postcard (tag-free binary)                                              | Protobuf (dag-pb)                        | [CBOR (deterministic subset with support for CIDs)](https://dasl.ing/drisl.html)      |
+| Hash                 | BLAKE3                                                                  | Configurable (SHA-256 default)           | Configurable via CID multihash                                                        |
```
vmx: DRISL only supports DASL-CIDs, hence only SHA-256.

docs/how-to/content-addressed-folders.md
```diff
@@ -0,0 +1,188 @@
...
+| Overhead             | ~0% (names + hashes only)                                               | 0.005–1% (depends on file count/size)    | Minimal (CBOR framing)                                                                |
+| Identifiers          | BLAKE3 hash (CID-encodable via `blake3` + `blake3_hashseq` multicodecs) | CID (self-describing)                    | CID (self-describing)                                                                 |
+| File lookup          | O(1) offset from root                                                   | DAG traversal, depth varies              | O(1) key lookup from root                                                             |
+| Subsetting           | Individual files only                                                   | Files and folders (subtree by CID)       | Individual files only                                                                 |
+| Byte ranges          | No (whole-file hashes)                                                  | Yes (chunked DAG allows partial reads)   | No (whole-file CIDs)                                                                  |
```
vmx: It isn't clear to me, what you mean with "Byte ranges". You can make range requests on the files. Subsets are addressed, implicitly due to Blake3.

2color: I'll just remove this, it's more about verifying byte ranges and there's a bit of nuance

- With blake3 (and by extension iroh collectons) you can do partial verification, but that requires getting the BAO outboard. 
- With UnixFS, that's natively possible,
- With DRISL/MASL that's not really possible

docs/how-to/content-addressed-folders.md
2color: No, they only allow one level of indirection.

docs/how-to/content-addressed-folders.md
```diff
@@ -0,0 +1,195 @@
...
+- **Deep DAG traversal.** Resolving a file means walking the directory DAG node by node — `a/b/c.csv` requires resolving `a`, then `b`, then the file, each a separate block fetch and decode. Nested directories and HAMT shards make the depth unpredictable.
+- **Mature ecosystem.** UnixFS has the broadest tooling support and is the de facto standard for IPFS content addressing, with implementations in Go ([Kubo](https://github.com/ipfs/kubo)) and TypeScript ([Helia](https://github.com/ipfs/helia)).
+
+## DASL: MASL and DRISL
+
+[DASL](https://dasl.ing) (Data Addressed Structures and Links) is a family of specs emerging from the Bluesky/AT Protocol ecosystem that provide content-addressed data structures built on CBOR rather than protobuf.
```
darobin: I don't think it's accurate to say that they're emerging from AT. The relationship to AT is mostly that we aligned with the subset of CIDs/IPLD/CAR they used (it wasn't the initial one we had) because alignment is good. But the other DASL specs aren't from AT (though some extend it).

docs/how-to/content-addressed-folders.md
```diff
@@ -0,0 +1,195 @@
...
+
+- Identical data always produces identical bytes (and therefore identical CIDs)
+- Native CID support via CBOR Tag 42
+- Strict constraints: string-only map keys, no indefinite-length arrays, restricted float representations
+- Each CID refers to one complete, discrete CBOR object
+
+**[MASL](https://dasl.ing/masl.html)** is a CBOR-based metadata system built on DRISL, designed for content-addressed and decentralized systems. It operates in two modes:
+
+- **Single mode** (`src`): wraps one resource with metadata (content type, etc.)
+- **Bundle mode** (`resources`): maps file paths to resource CIDs with per-file metadata — essentially a directory representation
```
darobin: Or maybe directory tree representation?

docs/how-to/content-addressed-folders.md
```diff
@@ -0,0 +1,195 @@
...
+
+Before comparing formats, it helps to understand what they all share: every approach effectively constructs a [_Merkle DAG_](../concepts/merkle-dag.md), a data structure which allows you to derive a small verification identifier like a CID to represents a collection of data.
+
+The formats below vary in how they construct the Merkle DAG and the trade-offs they make, but in essence they all allow you to produce a CID that represents a collection of files, such that verifying that hash verifies the entire contents.
+
+This matters for build outputs, software distributions, large datasets, website archives — any case where you need to verify that a collection of files hasn't changed.
+
+A naive approach like hashing a tarball is fragile: tar archives encode metadata (timestamps, permissions, ordering) that vary between machines, producing different hashes for identical file contents. It's also impractical for large datasets where you cannot afford to store two copies of the data.
+
+Content addressing solves this, but the choice of format has real consequences, particularly for overhead, determinism, language support and interoprability within an ecosystem.
```
bumblefudge:
```suggestion
Content addressing solves this, but the choice of format has real consequences, particularly for overhead, determinism, language support and interoperability within an ecosystem.
```

docs/how-to/content-addressed-folders.md
```diff
@@ -0,0 +1,195 @@
...
+- Native CID support via CBOR Tag 42
+- Strict constraints: string-only map keys, no indefinite-length arrays, restricted float representations
+- Each CID refers to one complete, discrete CBOR object
+
+**[MASL](https://dasl.ing/masl.html)** is a CBOR-based metadata system built on DRISL, designed for content-addressed and decentralized systems. It operates in two modes:
+
+- **Single mode** (`src`): wraps one resource with metadata (content type, etc.)
+- **Bundle mode** (`resources`): maps file paths to resource CIDs with per-file metadata — essentially a directory representation
+
+MASL bundles are conceptually similar to iroh collections: a flat map of paths to content hashes, no directory hierarchy nodes. The key difference is MASL also carries per-resource metadata (like content types) and uses CIDs (self-describing, multi-codec identifiers) rather than raw BLAKE3 hashes. Like iroh collections, subsetting operates at the individual file level — there is no native subdirectory addressing.
```
darobin: Don't know if you want to flag this but the per resource metadata is deliberately a set of HTTP headers. MASL is intended for web compatibility, to avoid all the nonsense about guessing media types and such.

docs/how-to/content-addressed-folders.md
```diff
@@ -0,0 +1,195 @@
...
+
+- [UnixFS](https://specs.ipfs.tech/unixfs/)
+- [iroh collections](https://docs.iroh.computer/protocols/blobs#collections)
+- [DASL](https://dasl.ing) along with its metadata system [MASL](https://dasl.ing/masl.html)
+
+## Merkle DAGs: the common foundation
+
+Before comparing formats, it helps to understand what they all share: every approach effectively constructs a [_Merkle DAG_](../concepts/merkle-dag.md), a data structure which allows you to derive a small verification identifier like a CID to represents a collection of data.
+
+The formats below vary in how they construct the Merkle DAG and the trade-offs they make, but in essence they all allow you to produce a CID that represents a collection of files, such that verifying that hash verifies the entire contents.
```
bumblefudge:
```suggestion
The formats below vary in how they construct the Merkle DAG and the trade-offs they make, but in essence they all allow you to produce a CID that represents a collection of files, such that verifying that hash against the particular style of DAG verifies the integrity of all of its contents.
```

docs/how-to/content-addressed-folders.md
```diff
@@ -0,0 +1,195 @@
...
+| Subsetting           | Individual files only                                                   | Files and folders (subtree by CID)       | Individual files only                                                                 |
+| Determinism          | By construction                                                         | Depends on DAG construction choices      | By construction (DRISL)                                                               |
+| Implementations      | Rust only                                                               | Go, JavaScript, Rust                     | Wide See [cross-implementation test suite](https://hyphacoop.github.io/dasl-testing/) |
+| IPFS Gateway support | No                                                                      | Yes                                      | Yes                                                                                   |
+| Ecosystem            | iroh/n0                                                                 | IPFS (broad)                             | AT Protocol/Bluesky                                                                   |
```
darobin: I wouldn't make the ecosystem just AT here, but up to you!

docs/how-to/content-addressed-folders.md
bumblefudge:
```suggestion
[DASL](https://dasl.ing) (Data Addressed Structures and Links) is a family of specs for decentralized-web building blocks, congruous with but slightly extending the CBOR serialization and DAG profile used by the Bluesky/AT Protocol ecosystem.
```

docs/how-to/content-addressed-folders.md
bumblefudge:
```suggestion
- **Bundle mode** (`resources`): maps file paths to resource CIDs with per-file metadata — essentially a logical unit representation, i.e. all the paths on a given domain apex (including nested ones)
```
not exactly a directory, can contain many directories included nested ones or none.  more like a map of [pathed] relrefs to resources.

docs/how-to/content-addressed-folders.md
bumblefudge:
```suggestion
MASL bundles are conceptually similar to iroh collections: a flat map of paths to content hashes, no directory hierarchy nodes. The key difference is MASL also carries per-resource metadata (like content types) and uses CIDs (self-describing, multi-codec identifiers) rather than raw BLAKE3 hashes. Like iroh collections, subsetting operates at the individual resource level — there is no native subdirectory addressing.
```

docs/how-to/content-addressed-folders.md
mishmosh: Yes, I think it's fair to say in use by the ATProtocol ecosystem, but didn't uniquely come from it. Maybe this?

```suggestion
[DASL](https://dasl.ing) (Data Addressed Structures and Links) is a set of simple, standard primitives for working with content-addressed, linked data. Designed as a web-friendly, interoperable subset of IPFS and IPLD primitives, DASL is used in production by the AT Protocol ecosystem, including Bluesky.
```

docs/how-to/content-addressed-folders.md
```diff
@@ -0,0 +1,195 @@
...
+Collection Hash = blake3(root blob)
+                = blake3(hash(meta) ‖ hash(file₁) ‖ hash(file₂) ‖ …)
+```
+
+These are standard BLAKE3 hashes, but they can be encoded as CIDs for interoperability with the broader content-addressed ecosystem. The [multicodec table](https://github.com/multiformats/multicodec/blob/master/table.csv) defines the necessary codes: hash function `blake3` (`0x1e`) and codec `blake3_hashseq` (`0x80`). This lets iroh collection hashes be referenced anywhere CIDs are used, without changing the underlying data.
+
+### Characteristics
+
+- **No metadata pollution.** Unlike tar/zip, there are no timestamps, permissions, or ownership fields. Two directories with identical file names and contents always produce the same hash, regardless of when or where they were produced.
+- **Positional, tag-free encoding.** Postcard serializes fields in declaration order with no field numbers or type tags. The `"CollectionV0."` magic header handles versioning.
```
bumblefudge: without having followed the Postcard docs link above, i would have no idea what the magic header is or how it handles versioning! link again and/or expand what/how that is

docs/how-to/content-addressed-folders.md
2color: I rewrote this paragraph to expand on the exact two properties.

docs/how-to/content-addressed-folders.md
2color: I edited it. See the latest version

PR #2285, opened 2026-05-29
docs: guide to replace public gateways with self-hosted IPFS

lidel:
New how-to for non-browser clients and web apps still calling ipfs.io or dweb.link, routing each case to a drop-in replacement:

- backend retrieve-only -> Rainbow, with trustless-only config for public hostnames
- backend retrieve and publish -> Kubo, plus pinning services or IPFS Cluster for redundancy
- browser JS -> @helia/verified-fetch, with self-hosted Someguy and Rainbow/Kubo to stay independent of public utilities

Linked from the IPFS Gateway sidebar group and the how-to index.

---

- Part of https://github.com/ipshipyard/roadmaps/issues/53

cc @aschmahmann @achingbrain @mishmosh @bumblefudge


lidel:
I'm going to merge this as-is, feel free to refine in follow-up PRs, but we need these docs asap.

